PR #9, opened 2012-01-12
Fixed nosetest issue in create_stream

tgiguere:
Fixed a problem with create_stream causing unit test to fail.


tgiguere:
Closing this request because an extra file got commited.

PR #13, opened 2012-01-13
Fixed interface

dstuebe:
Fixed a number of interface issues. Only not implemented errors left in the validation log.


dstuebe:
can not merge... sigh.

PR #31, opened 2012-01-30
Changed AT to PRED in ion/services/dm

swarbhanu:


dstuebe:
Reviewed and made some changes. will make a new PR from ooici-dm

PR #33, opened 2012-02-03
Blog

dstuebe:
Dm sprint to implement elaboration capabilities using simple blog data.


mmeisinger:
ok, looks good at first glance. Go ahead and merge it it all works
Suggest renaming replay_agent to replay_process (or worker), but this can be a discussion item

dstuebe:
Yes - will sort out these issues next week.

PR #38, opened 2012-02-07
Luke: (Replay)

lukecampbell:
Changed the logic in replay. Replay now processes one post_id (soon to be dataset_id) and traverses a one-tier association and returns all the data and streams it. 
- Added new utility transform: TransformCapture (dumps stream output to file)
- Changed Replay to Standalone Process

NOTE: THIS PULL REQUEST SHOULD BE MERGED AFTER ION-DEFINITIONS HAS BEEN MERGED WITH https://github.com/ooici/ion-definitions/pull/69/files


mmeisinger:
looks good. please merge manually with proper ion-definitions in coi-services

lukecampbell:
Done deal

ion/services/dm/inventory/data_retriever_service.py
```diff
@@ -3,11 +3,12 @@
...
 @description Data Retriever Service
 '''
-from interface.services.dm.ireplay_agent import ReplayAgentClient
+
 from interface.services.dm.idata_retriever_service import BaseDataRetrieverService
+from interface.services.dm.ireplay_process import ReplayProcessClient
 from interface.objects import Replay
 from pyon.public import PRED
-from pyon.service.service import BaseService
+
```
lukecampbell, line 12: in on_start I manually attach a publisher stream for every stream_name passed in as 'publish_streams'.  Normally this is done by _set_publisher_endpoints in procs.py but this is only done for stream_processes, is there a way we could add a conditional to _spawn_standalone or maybe another type that calls _set_publisher_endpoints if publish_streams is passed in with as a config arg?

daf:
I don't know why this can't be auto-merged - it's just a normal fast forward for me if you do the manual merge as instructed.  @mmeisinger if you can do the same, it should be ok.

ion/services/dm/inventory/data_retriever_service.py
```diff
@@ -3,11 +3,12 @@
...
```
lukecampbell, line 11: in on_start I manually attach a publisher stream for every stream_name passed in as 'publish_streams'. Normally this is done by _set_publisher_endpoints in procs.py but this is only done for stream_processes, is there a way we could add a conditional to _spawn_standalone or maybe another type that calls _set_publisher_endpoints if publish_streams is passed in with as a config arg?

mmeisinger:
merged

ion/services/dm/inventory/data_retriever_service.py
mmeisinger, line 11: we can add the addition of publishers based on config for other process types too, if this seems to be a generic boilerplate reducing pattern.

PR #40, opened 2012-02-08
Sysname

dstuebe:
Fixed scoping for Publisher and Subscriber to use sysname. Repaired DM integration tests.


mmeisinger:
Looks good at quick glance

PR #52, opened 2012-02-13
Luke: Patched int test for dataset management

lukecampbell:
- Added support for the new data container.
- Added statistical checking (using normal distribution) to ensure
  quality and accurate map/reduce queries for bounds checking.


ion/services/dm/inventory/test/dataset_management_test.py
```diff
@@ -5,12 +5,13 @@
 '''
 from interface.services.dm.idataset_management_service import DatasetManagementServiceClient
 from ion.services.dm.inventory.dataset_management_service import DatasetManagementService
-from interface.objects import DataSet
+from prototype.sci_data.ctd_stream import *
```
dstuebe: Don't import *

PR #53, opened 2012-02-13
Luke: (Dataset Management) INT Tests

lukecampbell:
- Patched integration tests
- Patched DSM to use stream_ids when querying.


dstuebe:
Looks good.

PR #55, opened 2012-02-13
Luke: Data retriever service

lukecampbell:
- Added field to define_replay for datastore_name
- Added a patch in ctd_stream_publisher to track and kill spawned greenlets.


mmeisinger:
see comment on ion-definitions push.
The datastore name should be inferred from knowing the dataset id, not as an argument here.
The dataset need to reference the datastore.

mmeisinger:
pls manually merge with ion-definitions properly set when ready.

dstuebe:
Okay - thanks Michael

PR #56, opened 2012-02-13
Luke: Dataset Refactoring

lukecampbell:
- Refactored DM classes to support datastore_name field in datasets, good call. 
- Now data retriever, replay and dataset mgmt will have necessary knowledge about datastores and datasets.


dstuebe:
Here we go...

PR #61, opened 2012-02-14
Ingest update

dstuebe:
Made corrections to ingest service interface and added database profile as an argument to set the couch designs.


dstuebe:
Reviewed with the DM team...

PR #62, opened 2012-02-14
Luke: Refactoring

lukecampbell:
- Added ingestion launching rel file
- Refactored Replay and Dataset Retriever to use datasets in lieu of ad-hoc queries.


dstuebe:
Great work luke - the retriever/dset mgmt are really coming together.

PR #64, opened 2012-02-14
Ingestion delete

swarbhanu:
Patched delete_ingestion_configuration so that associations and transforms that were associated to the configuration are deleted also.


dstuebe:
Hi Swarbhanu
please have a look at the changes I made. It is important that we never except: or except Exception: That does nothing but bury the error.
David

PR #71, opened 2012-02-15
Luke: Replay Process

lukecampbell:
- In lieu of patching streams in the process, we opt to use the built-in
  mechanics of the container, thanks Michael!


dstuebe:
Great

PR #72, opened 2012-02-15
Luke: CEI Integration into DM Services

lukecampbell:


mmeisinger:
very nice, great stuff

dstuebe:
Had some trouble with the merge - I think there are still a few issues. Since the process_definition is now created in on_start for ingestion mgmt svc and Retriever svc it is not present when the svc class is instantiated for unit tests.
We need to add a mock self.process_definition object for these tests.
See my latest changes here...
https://github.com/ooici-dm/coi-services/tree/cei_integration

dstuebe:
Merge madness!

PR #73, opened 2012-02-16
Stream def

dstuebe:
Fixed pubsub interface to use stream def


dstuebe:
Will have to fix this later...

PR #74, opened 2012-02-16
Luke: (Replay) Added chopping

lukecampbell:


dstuebe:
It is a good strategy for what we have at the moment. We need to revisit it later. The intent is to chop in a way that sends a useful block of information - ie the same granule with less content. That drives up the metadata/data ratio and the waste in the stream, but what we hope to have is content based routing of packets - so each packet must contain something useful - by itself.

PR #75, opened 2012-02-16
Process Dispatcher bridge service

labisso:
Added an implementation of PD that either 1) spawns processes in the local container or 2) spawns processes with the external CEI Process Dispatcher, depending on whether it is started via CEI launch plan. Further explanation is in code comments.

This is my first foray into Pyon service development, so I would appreciate a quick code review. Note that this branch needs to be merged at the same time as ooici/ion-definitions#93. Once approved I wil perform both merges.


ion/services/cei/process_dispatcher_service.py
```diff
@@ -77,7 +126,7 @@ def dissociate_execution_engine(self, process_definition_id='', execution_engine
                                                           execution_engine_definition_id)
         self.clients.resource_registry.delete_association(assoc)
 
-    def schedule_process(self, process_definition_id='', schedule=None):
+    def schedule_process(self, process_definition_id='', schedule=None, configuration={}):
```
daf, line 129: Dont' use `={}` - use `=None` and set it in your method body.  Seems you are already doing so around line 149.

Mutable kwargs are bad news!

labisso, line 129: Yeah I noticed that, but that is how the generated service interface looks. It seems to be a common pattern in the service YAML to have empty dicts for default parameter values, which I don't like. Better to use `None` across the board there?

ion/services/cei/process_dispatcher_service.py
```diff
@@ -3,11 +3,60 @@
...
+    # 2. Otherwise, processes are started directly in the local container.
+    #    This mode is meant to support the developer and integration use of
+    #    r2deploy.yml and other single-container test deployments.
+    #
+    # Note that this is a relatively short-term situation. The PD will soon
+    # natively run in the container and these tricks will be unnecessary.
+
+    def on_init(self):
+
+        #am I crazy or does self.CFG.get() not work?
```
mmeisinger, line 42: self.CFG.get() should work, because it's a (modified) dict, but there is always potential for a bug.

mmeisinger:
looks good to me. thanks. please merge once tests pass. I've already merged the ion-definitions

It would be nice to add a link to the repo where the code for the PD implementation is

PR #76, opened 2012-02-16
Luke: Replay Patched Nonetype bug

lukecampbell:
- Default Args are now NoneType and not {} resulting in small bugs,
  patched one.


mmeisinger:
good. you can also use self.CFG.get_safe("xxx.yyy", {}) to avoid some of the clutter

PR #78, opened 2012-02-18
Hdf test

dstuebe:
Add hdf and numpy as dependencies.
Nothing will fail without out the dependencies - yet.


dstuebe:
Let Jamie merge this pull request - not late friday afternoon!

mmeisinger:
go when Jamie is ready

PR #103, opened 2012-03-20
Bug fixes and mods requested by Michael

dstuebe:
- Added dataset id to dataset config object and granule ingested event.
- Fixed bugs in last update worker
- Added transform_id to the config when a transform is launched
- Set origin of ExceptionEvent to be the transform id when sent by transform worker
- Fixed issues in unit tests caused by changes.


swarbhanu:
Hi Patrick,
CC: David LaBissoniere, David Stuebe,

When I run services on launch_plan (for example: r2dm.yml) and a level fails because of a failed service, where would I find the detailed log for the failed service so that I can find what exactly went wrong in that service?

I wanted to know if it is possible to see the error messages the way we can when we run the services in a container.

Thanks,
Swarbhanu.

PR #105, opened 2012-03-26
Webui full

lukecampbell:


dstuebe:
Here is a great first go at adding visual navigation of resources and association. Visualization of graph structures rather than just trees is on the way.

PR #107, opened 2012-04-02
Luke: Updated README for matplotlib on Lion

lukecampbell:


lukecampbell:
This URL has also helped a few 10.6 installations as well

PR #110, opened 2012-04-19
Process Dispatcher emits ION events

labisso:
Adds events to the Process Dispatcher. Clients should subscribe to events for a process before requesting it to be scheduled. The following events can be emitted (from `ProcessStateEnum`):
- `SPAWN` - when a scheduled process is running in a container (somewhere)
- `TERMINATE` - when a process either dies on its own or is killed on request
- `ERROR` - when a process fails to start or dies abnormally

To launch processes, the following behavior is recommended:

``` python
# first call create_process to get the process id
pid = self.pd_cli.create_process(process_definition_id)

# next create an event subscriber for this process
event_sub = EventSubscriber(event_type="ProcessLifecycleEvent",
    callback=self._event_callback, origin=pid,
    origin_type="DispatchedProcess")

# then request the process to be scheduled
self.pd_cli.schedule_process(process_definition_id, process_id=pid)
```

Once scheduled, wait for appropriate (asynchronous) events before trying to use the process. If successful, you will get a `ProcessStateEnum.SPAWN`.

I'm open to suggestions for improving the use of events. This code still needs to be tested under the CEI lightweight launch.

### DIFF
--- a/ion/services/cei/process_dispatcher_service.py
+++ b/ion/services/cei/process_dispatcher_service.py
@@ -6,11 +6,15 @@ __license__ = 'Apache 2.0'
 import uuid
 import json
 
+import gevent
+
 from pyon.public import log, PRED
 from pyon.core.exception import NotFound, BadRequest
 from pyon.util.containers import create_valid_identifier
+from pyon.event.event import EventPublisher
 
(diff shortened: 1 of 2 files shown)


dstuebe:
Hi David
This looks like a great start and a great pull request - thanks for the time spent explaining how to use the new code.

David

PR #112, opened 2012-05-04
EOI Dataset Agent Framework realignment

blazetopher:
@MauriceManning
Core realignment of EOI Dataset Agent with the MI Instrument Agent


mmeisinger:
Very good pregoress. Strong towards alignment.

The only nit I have is package naming. I would place the agent in ion.agents.data or something like that (instead of eoi). Subsystem boundaries get blurrier and blurrier over time. It's more the primary responsibility. E.g. the MI team may use and extend a Dataset agent to get data from a glider platform in. I know for services we use subsystem packets, but I even don't like it there (e.g. one subsystem creating a service and another subsystem adding an operation to the existing service). Having strong subsystem boundaries leads to more stove-piped thinking than necessary.

We'll still have to play a bit with moving things into the right places. No immediate action to take unless you are just doing it anyways.

PR #114, opened 2012-05-08
Actor refactor

thomaslennan:
Attempted to find all the touch points where UserIdentity and UserInfo were used and refactored to ActorIdentity.


mmeisinger:
See other PRs.

In general the changes are OK, but I do not agree with removing the UserInfo resource. Please reintroduce the UserInfo with service operations as before. Two ActorIdentity can associate with the same UserInfo

PR #115, opened 2012-05-08
UserIdentity refactor

thomaslennan:


mmeisinger:
Very good refactor. Thanks for the flexibility.

I suggest merge when ready

PR #119, opened 2012-05-15
Change name

dstuebe:
Remove the word 'CTD' from the example code using the generic granule and publisher infrastructure because it causes confusion.


MauriceManning:
Can we refactor the SimpleCtdPublisher base class as well?

dstuebe:
Sorry - did not get back to this yesterday - do you still want me to make that change? There are only a dozen or so places where it is used in the main line right now...

PR #123, opened 2012-05-17
Luke: adds read_mult to resource registry service

lukecampbell:


mmeisinger:
Looks good to me. Suggest manual merge with needed ion-defs.

One nit: Please place the crud operations before the find and special operations in the class. read_mult should be directly below read, not below find. Just tidiness of the code.

PR #129, opened 2012-05-29
Patch for ingestion management

lukecampbell:
Ingestion Management relies on the bootstrap for initializing the resource for the process definition for ingestion workers.  The bootstrap is initialized in most of the deployment rel files but in bare containers or custom rel files it is not a guarantee that the bootstrap was launched and that the resource does in fact exist.

These changes make checks against the existence of the resource and in a call to `create_ingestion_configuration` the existence is checked again and if it does not exist, the resource is then created.

### DIFF
--- a/ion/services/dm/ingestion/ingestion_management_service.py
+++ b/ion/services/dm/ingestion/ingestion_management_service.py
@@ -51,6 +51,7 @@ class IngestionManagementService(BaseIngestionManagementService):
             raise StandardError('Invalid CFG for core_xps.science_data: "%s"; must have "xs.xp" structure' % xs_dot_xp)
 
         self.serializer = IonObjectSerializer()
+        self.process_definition_id = None
 
 
     def on_start(self):
@@ -61,7 +62,8 @@ class IngestionManagementService(BaseIngestionManagementService):
             restype=RT.ProcessDefinition,
             name='ingestion_worker_process',
             id_only=True)
-        self.process_definition_id = res_list[0]
+        if len(res_list):
+            self.process_definition_id = res_list[0]
 
 
     def on_quit(self):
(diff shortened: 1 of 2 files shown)


ion/services/dm/ingestion/ingestion_management_service.py
```diff
@@ -61,7 +62,8 @@ def on_start(self):
             restype=RT.ProcessDefinition,
             name='ingestion_worker_process',
             id_only=True)
-        self.process_definition_id = res_list[0]
+        if len(res_list):
+            self.process_definition_id = res_list[0]
```
lukecampbell, line 66: The `res_list` may not have elements in it if the bootstrap service is not started.  Bootstrap is not a requirement in the service clients list therefore we can't make assumptions about the existence of resources it creates.  If bootstrap service was not launched and line 64 is executed the container will crash due to a `IndexError: list index out of range`.

ion/services/dm/ingestion/ingestion_management_service.py
```diff
@@ -77,6 +79,12 @@ def create_ingestion_configuration(self, exchange_point_id='', couch_storage=Non
         @param number_of_workers is the number of ingestion workers to create
         """
 
+        if self.process_definition_id is None:
+            process_definition = ProcessDefinition(name='ingestion_worker_process', description='Worker transform process for ingestion of datasets')
+            process_definition.executable['module']='ion.processes.data.ingestion.ingestion_worker'
+            process_definition.executable['class'] = 'IngestionWorker'
+            self.process_definition_id = self.clients.process_dispatcher.create_process_definition(process_definition=process_definition)
+
```
lukecampbell, line 87: If the service has been started but `self.process_definition_id` has still not been initialized then bootstrap has not been run and the resource needs to be created before operations can continue successfully.

PR #130, opened 2012-05-30
Add support for Pyon Execution Engine Agent

oldpatricka:
This patch set adds support for the execution engine agent.


mmeisinger:
Very nice. Looks good to go

I think the commands should be implemented as device/resource commands with a rcmd_ prefix and called via execute(). The reason is that agent commands target the agent process itself, while resource commands target the resource that is managed by the agent (here: the execution engine = the container). There should be very few to none agent commands.

oldpatricka:
Thanks Michael, I just switched to rcmd_.

PR #132, opened 2012-05-30
Preservation

dstuebe:
Do not merge - this is for comparison and discussion.

See: https://docs.google.com/drawings/d/1Yg1ibe-TlIkjb29CHInpxq5jXEkIkt9cEX1iReUzKgk/edit?pli=1

This is a basic implementation of Resource Management for Preservation. It does not actually 'do' anything yet. See prototypes for changes to pyon FileSystem which provide mkfs, mount and umount commands.


mmeisinger:
Looks good to me. No issues against merging if everything passes.

We should complement this with preload or bootstrap logic to actually create the instances that we intend to run the system against.

PR #134, opened 2012-05-31
C1 notify

dstuebe:
Cleaned up UNS and generalized event processor to support email, sms and event detection.


mmeisinger:
Looks good on first glance. Substantial amount of work.

PR #139, opened 2012-06-06
Luke: really annoying bug

lukecampbell:


lukecampbell:
Goes with the test bootstrap

PR #140, opened 2012-06-06
Geo searching

lukecampbell:


dstuebe:
Glad we have a basic capability in place for geo search. Limited to point objects for now. Thanks Luke

PR #152, opened 2012-06-14
Luke: association deletion

lukecampbell:


dstuebe:
Hmmm - I don't think this is quiet right yet.
For instance with the index - the delete index function should not delete the associations from a catalog created by catalog mgmt service. It is the responsibility of the user to delete the catalogs association (or the entire catalog) before deleting the index. That pattern should be implemented in the tests as well so that associations are not left behind and the correct service which is responsible for the association is deleting it.

dstuebe:
This looks better. Lets see if all the warnings go away.

lukecampbell:
Where are the warnings?

dstuebe:
Here is what to look for:
https://github.com/ooici/pyon/blob/master/pyon/datastore/couchdb/couchdb_datastore.py#L308

David

On Thu, Jun 14, 2012 at 11:36 AM, Luke Campbell <
reply@reply.github.com

PR #164, opened 2012-06-28
Luke: fixes discovery query problems (interim)

lukecampbell:


lukecampbell:
Still need to figure out how to properly index and map Identifiers to un-break this feature of searching.  ElasticSearch currently uses the CouchDB document id as it's document ids, this breaks searching based on document ids which does not appear to have broken additional code.

PR #167, opened 2012-07-02
Finishes basic implementation of primitive dm end to end

lukecampbell:
This is the more concrete implementation and basic end-to-end implementation for the primitive dm work. 
I changed `retrieve` to utilize the service's instance to process the data instead of spawning a process and using RPC.  It should behave like `execute_transform`.  Once TMS work is done, I'll refactor it to extend the `execute_transform` interface. 

I addressed the bug with deserialization in the example data producer by simply un-nesting the record dictionaries.

Previously the example data producer nested the record dictionary tools inside the the array which `numpy` handles fine and `msgpack` can serialize and deserialize naturally. However, in our interceptor stack it sees an IonObject and tries to deserialize the nested objects when the granule is sent through RPC requests and returns.  I attempted a primitive deserializer but it's not an easy fix because all the attributes for nested objects would need to be stripped and analyzed to see if the interceptor stack will deserialize them.  It would be slow and ugly so I opted for the easier solution, fix the example.

Next is unit and integration testing.


ion/processes/data/example_data_producer.py
```diff
@@ -80,30 +80,25 @@ def _trigger_func(self, stream_id):
...
 
-            rdt1['time'] = tvar
-            rdt1['lat'] = lat
-            rdt1['lon'] = lon
-
-            rdt['group1'] = rdt1
-            rdt['group0'] = rdt0
+            rdt['time'] = tvar
+            rdt['lat'] = lat
+            rdt['lon'] = lon
```
lukecampbell, line 101: This change was necessary and I need to illustrate to everyone, we CAN NOT embed python objects into granules, don't try it we're not there yet. The decoder can't handle it because of conflicts with `msgpack` and `numpy`.  I tried writing a decoder but the problem runs deeper than I thought. **Granules should be monolithic and not nested.**

ion/processes/data/replay/replay_process.py
```diff
@@ -43,8 +43,8 @@ def on_start(self):
 
         self.dataset_id      = self.CFG.get_safe('process.dataset_id', None)
         self.delivery_format = self.CFG.get_safe('process.delivery_format',{})
-        self.start_time      = self.CFG.get_safe('process.delivery_format.start_time', None)
-        self.end_time        = self.CFG.get_safe('process.delivery_format.end_time', None)
+        self.start_time      = self.CFG.get_safe('process.query.start_time', None)
+        self.end_time        = self.CFG.get_safe('process.query.end_time', None)
         self.publishing      = Event()
```
lukecampbell, line 48: Moved start/end time out of delivery format params and into query.

ion/services/dm/inventory/data_retriever_service.py
```diff
@@ -128,14 +130,26 @@ def cancel_replay(self, replay_id=''):
...
+
+        replay_instance.dataset = self.clients.dataset_management.read_dataset(dataset_id)
+        replay_instance.start_time = query.get('start_time', None)
+        replay_instance.end_time = query.get('end_time', None)
+        replay_instance.container = self.container
+
+        retrieve_data = replay_instance.execute_retrieve()
 
-        return cli.execute_retrieve()
+        return retrieve_data
```
lukecampbell, line 153: This entire method is now run in the instance in lieu of starting a process.  I will add some data length checking soon.

PR #174, opened 2012-07-03
Luke: more fixes

lukecampbell:


ion/core/bootstrap_service.py
```diff
@@ -292,7 +292,7 @@ def restart_transform(transform_id):
                 restart_transform(transform_id)
 
     def post_index_creation(self,config):
-        if self.CFG.get_safe('system.elasticsearch'):
+        if self.CFG.get_safe('system.elasticsearch') and self.CFG.get_safe('bootstrap.use_es'):
```
lukecampbell, line 295: Index creation must explicitly be configured.

Use `bin/pycc --rel res/deploy/r2deploy.yml bootstrap.use_es`

PR #175, opened 2012-07-05
More tests and skips discovery tests for --with-pycc

lukecampbell:


MauriceManning:
The list func is useful but how will I know 'which' ingest config is my sci_data vs my raw data? Just using conventions? Should I be able to leverage the taxonomy info to direct me?

lukecampbell:
well ingestion config is per exchange point.  So you would inspect the ingestion configs to see if the exchange point matches what you're looking for.  If you need further inspection you may inspect each queue and see if the queue type matches your data.  Eventually queue type will be something like an xpath for the data that is transmitted on streams, for now it's a string.

lukecampbell:
Something like

```
[i for i in ingestion_management.list_ingestion_configurations() if i.exchange_point == 'science_data']
```

PR #177, opened 2012-07-06
Luke: Refactor in retrieve (ready for merge)

lukecampbell:
Retrieve was refactored from `start_retrieve`.  I patched and added better support for domain variable slicing via `time`.  We don't have a specific set of taxonomies or stream definitions yet so it is hard-coded as `time` which is till really intuitive.  The default if `time` is not a field in the record dictionary is to use the time when the granule arrived.

### DIFF
--- a/ion/processes/data/ingestion/science_granule_ingestion_worker.py
+++ b/ion/processes/data/ingestion/science_granule_ingestion_worker.py
@@ -9,10 +9,11 @@ from pyon.core.bootstrap import get_sys_name
 from pyon.core.interceptor.encode import encode_ion
 from pyon.core.object import ion_serializer
 from pyon.ion.process import SimpleProcess
+from pyon.ion.granule import RecordDictionaryTool
 from pyon.net.endpoint import Subscriber
 from pyon.datastore.datastore import DataStore
 from pyon.util.arg_check import validate_is_instance
-from pyon.util.containers import get_ion_ts
+from pyon.util.containers import get_ion_ts, get_safe
 from pyon.util.file_sys import FileSystem, FS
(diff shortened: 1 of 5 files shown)


ion/processes/data/ingestion/science_granule_ingestion_worker.py
```diff
@@ -46,9 +48,13 @@ def ingest(self, msg, stream_id):
...
-
-
-
+        rdt = RecordDictionaryTool.load_from_granule(msg)
+        time = get_safe(rdt,'time')
+        if time is not None and isinstance(time,numpy.ndarray):
+            time = time[0]
+        else:
+            time = None
+
```
lukecampbell, line 57: We don't have a standardized set of taxonomies _yet_ so I'm making the assumption that there should be a field called `time`. If there is not a field and it's not filled with a numpy array then a generic timestamp is used.

ion/services/dm/test/test_dm_end_2_end.py
```diff
@@ -118,7 +173,28 @@ def test_dm_end_2_end(self):
...
+
+        config_id = self.get_ingestion_config()
+
+        dataset_id = self.ingestion_management.persist_data_stream(stream_id=stream_id, ingestion_configuration_id=config_id)
+
+        self.publish_fake_data(stream_id)
+
+        self.wait_until_we_have_enough_granules(dataset_id,2) # I just need two
+
+        replay_granule = self.data_retriever.retrieve(dataset_id,{'start_time':0,'end_time':2})
```
lukecampbell, line 191: This is an example of specifying the domain of the granule to be returned.

PR #179, opened 2012-07-11
Luke: Adds visual map utility to web utility

lukecampbell:
`http://localhost:8080/map` displays a google-map with all resources containing geometric points and contains simple click events for creating new resources (great for testing and demos).

To use launch your capability container with the elasticsearch bootstrap

```
bin/pycc -fc --mx --rel res/deploy/r2deploy.yml bootstrap.use_es=True
```

Also adds support for URL args being passed into a new-resource action for specifying the attributes. 

I'm only adding it to show some neat capabilities, it is not intended to be used as is for production level code and will not be maintained thoroughly.


mmeisinger:
Looks good. Impressive tech demo.

PR #188, opened 2012-07-17
UNS and Notification workers

swarbhanu:
Changes include all major UNS and notification worker functionality required for this iteration.

A summary of the main changes in the UNS code base:
- Refactored  UNS - Deprecated SMS. SMS is only a matter of sending email to the sms provider
- Using CEI to create notification workers
- Each notification worker and also the UNS maintains a dictionary holding all the user info including the notifications he is interested in
- Notification workers subscribe to special event, ReloadUserInfoEvent, that occurs every time the UNS is used to change something about the notifications using its CRUD methods
- The UNS updates its dictionary, user_info, automatically without listening to events. We can do this assuming that only the UNS is used to change anything to do with notifications.
- Aggregate notifications (batch notifications) are processed via UNS
- Notification workers handle immediate email response for notification events.
- Using a fake scheduler, we have implemented a publish_event method that publishes an event at a future time
- Tests concerning UNS and the workers have been implemented for all major functionality.
- create_email method of UNS which was earlier used to set up a specific email notification for a single user and event type has been deprecated (commented out). This is because the rest of the existing set up in UNS is enough and very easy to use as shown in the test code. We dont need a separate create_email() method.

For coi-services, these are the four files that have been modified/created:
1. ion/services/dm/presentation/user_notification_service.py
2. ion/processes/data/transforms/notification_worker.py
3. ion/services/dm/utility/uns_utility_methods.py (where all common utility methods are kept. These methods are used both by the workers and the user notification service)
4. ion/services/dm/presentation/test/user_notification_test.py

For ion-definitions, the changes are in only these three files: 
1. Ion-definitions/object/data/dm/notification.yml
2. Ion-definitions/objects/services/dm/user_notification_service.yml
3. Ion-definitions/objects/data/event.yml

New event type such as the ReloadUserInfoEvent and the SchedulerEvent (used by the fake scheduler) were created in data/events.yml. Please checkout the latest in the ion-definitions (please refer to https://github.com/swarbhanu/ion-definitions/tree/batch_process_notifications).


swarbhanu:
This requires ion-defs, to point to the commit 73795e3df89f39b9eb302203. I have a separate pull request on ooici/ion-definitions. The above commit is in that pull request.

swarbhanu:
The latest commits require ion-defs, to point to the commit e331106 which is in a separate pull request on ooici/ion-definitions.

PR #198, opened 2012-07-24
Simple implementation new transform tests

swarbhanu:
A simple implementation of two INT tests for the new transforms:
1. Listen for events of a certain type and if 3 events are received within a certain space of time, publish an alert event.
2. Listen for streams and if certain parameters are a match for the stream, publish an alert event.

These commits should not break anyone else's code.


swarbhanu:
Since I wrote this code on top of the new User Notification Service code, it will be better to merge the pull request related to the User Notification Service before merging this pull request.

PR #204, opened 2012-07-31
Using parameter dictionary instead of taxy tool and taxonomy

swarbhanu:
Using parameter dictionary instead of taxy tool and taxonomy in ctd_transforms and the test for ctd_transforms.


lukecampbell:
Please don't merge this yet.

PR #215, opened 2012-08-02
Using a real scheduler service to publish an event at a particular time

swarbhanu:
The caller must create a timer object of type, SchedulerEntry, using the scheduler service. The caller then passes in the event and the timer object to the publish_event() method of UNS. The UNS will publish the event at the correct time (or times) depending on whether the timer object is a time of day or an interval kind of timer. The test for the publish_event() method checks out this functionality.


swarbhanu:
There is a pull request in ion-definitions (from my own repository, branch = user_notific_scheduler). It has to be merged first before this pull request can be automatically merged.

PR #221, opened 2012-08-03
This one should definitely fix the smoke tests.

swarbhanu:


ion/services/ans/workflow_management_service.py
```diff
@@ -168,8 +169,21 @@ def create_data_process_workflow(self, workflow_definition_id='', input_data_pro
                 data_product_name = create_unique_identifier(workflow_definition.name + '_' + data_process_definition.name)
 
             # Create the output data product of the transform
-            transform_dp_obj = IonObject(RT.DataProduct, name=data_product_name,description=data_process_definition.description)
-            transform_dp_id = self.clients.data_product_management.create_data_product(transform_dp_obj, process_output_stream_def_id)
+
+            craft = CoverageCraft
```
daf, line 173: Shouldn't this just be `CoverageCraft()`

ion/services/ans/test/test_helper.py
```diff
@@ -32,20 +45,44 @@ def create_ctd_input_stream_and_data_product(self, data_product_name='ctd_parsed
...
 
 
         log.debug('Creating new CDM data product with a stream definition')
-        dp_obj = IonObject(RT.DataProduct,name=data_product_name,description='ctd stream test')
-        try:
-            ctd_parsed_data_product_id = self.dataproductclient.create_data_product(dp_obj, ctd_stream_def_id)
-        except Exception as ex:
-            self.fail("failed to create new data product: %s" %ex)
+
+        craft = CoverageCraft
```
daf, line 72: Same, `CoverageCraft()`

swarbhanu, line 72: Yeah, I think the line is okay.

craft = CoverageCraft
        sdom, tdom = craft.create_domains()
        sdom = sdom.dump()
....
And so on..
- Swarbhanu.

On 8/3/12 2:46 PM, "Dave Foster"
<reply+p-1304407-13d6a3c569cc07c860d61dbf83c8ad040b8a9f8d-1181645@reply.git
hub.com> wrote:

daf, line 72: Ah I see, class methods.  Still looks a bit funny, but I guess it's fine.

swarbhanu, line 72: Yep - Swarbhanu.

On 8/3/12 2:58 PM, "Dave Foster"
<reply+p-1304542-ae27b9a29bc45e63337cada15b5f3d72ceb12b92-1181645@reply.git
hub.com> wrote:

swarbhanu, line 72: Hi Maurice,

Dave F. solved the problem in the SimpleCtdDataProducer. It was to do with the use of a while True loop in the data producer. So even when the processes were being shutdown, this loop was still running and at some point it found that its publisher was sort of dying or not there anymore.

Dave F. used a pattern which sets a gevent event when the process is closing down and then uses this event in the while loop in order to make sure that when the process is closing down, the loop ends right there. It is a great pattern. It definitely solved the error I was seeing in the SimpleCtdDataProducer.

There is a pull request, PR 228, which contains the patch fix.

As I mentioned to Dave, there are several modules in coi-services which are similar in what they want to accomplish (publish data streams or publish granule within while loops) which use the "while True" pattern, which as we have found out is not safe. Some of these modules are actually not being used in tests but when they do get used, they will throw the above Exception, which can become tricky to pin point.

As Dave mentioned, may be someone (could be me) can replace the "while True" with Dave's pattern during design week as part of code clean up.

I think that modules that do their work at the top level should not really be using the "while True" a lot but use the above pattern because in reality they are not meant to run forever and so they need to shut down when they should shut down and it is better if the programmer knows when it is supposed to shut down.

Thanks,
Swarbhanu.

swarbhanu, line 72: Hi Jamie,
CC: Dave,

I am copying Dave in this conversation because I think the test failure has to do with how big couchdb clusters work and the pyon datastore, and Dave might correct me if I am wrong.

Regarding the failing tests in UNS in the big couchdb cluster test (coi_testbc), I think this is what happened. I was using events in the test that was of a very generic type, ResourceLifeCycleEvent and an origin (that may have been popularly used in other test modules). I think that in the big couchdb cluster, the events are not cleaned up in the datastore as soon as a test module runs its course, and so when another test tries to use the datastore to search for events it sometimes encounter many residual events (that are hanging around in the couchdb datastore even after those other tests have ended).... Both the tests that failed were using my find_events() method in UNS and this method just uses the datastore's query_view() method.

The relevant code is here:
https://github.com/ooici/coi-services/blob/master/ion/services/dm/presentation/user_notification_service.py#L563-620

Therefore, my test failed saying that it found 151 events instead of the 4 that it is supposed to find (which it does find in normal testing). My tests pass in a normal test:
ex: bin/nosetests ion/services/dm/presentation/test/testuser_notification_service.py

I am confident that if the test was rerun it would show a different number of events found.

So what I have done now is use only very unique event types and more importantly, origin, in my tests. I think this will solve the problem for my test. However, I think that later on it may be good to see whether cleanups of the datastores are occurring in a timely manner after each test is completed in the "couchdb big cluster" test.

I am merging in my changes in coi-services. Please let me know if the tests fail again.

Unfortunately, I am not on a VPN and as far as I know our asascience network has strong firewalls and will not allow me to access the OOICI vpn. It will be great if you can let me know if the tests fail again.

Thanks,
Swarbhanu.

ion/services/ans/test/test_helper.py
```diff
@@ -32,20 +45,44 @@ def create_ctd_input_stream_and_data_product(self, data_product_name='ctd_parsed
...
+        sdom, tdom = craft.create_domains()
```
lukecampbell, line 73: Please get rid of these and use `time_series_domain`.  Example:

``` python
from ion.services.dm.utility.granule_utils import time_series_domain
tdom, sdom = time_series_domain()
```

ion/services/ans/workflow_management_service.py
lukecampbell, line 173: Please get rid of these and use `time_series_domain`.  Example:

``` python
from ion.services.dm.utility.granule_utils import time_series_domain
tdom, sdom = time_series_domain()
```

ion/services/ans/test/test_helper.py
swarbhanu, line 72: Hi Maurice,
CC: Luke, Dave, Tim, Carlos,

I have added the replace_data_process() method to add replace functionality to data process management service.

The ion def pull request is here: https://github.com/ooici/ion-definitions/pull/262/files

The coi-services pull request : https://github.com/ooici/coi-services/pull/495/files

Please review the pull requests and let me know if you would like me to make any changes.
If after review, the pull requests are found to be okay for merging, the ion defs pull request needs to be merged first.

Thanks,
Swarbhanu.

Swarbahnu
1. Can you verify that our data process and data process definition tests demonstrate the remove op? I think this is another place where we may want to augment the service interface to provide the correct sequence of steps that send events,  disconnect from the input stream, delete the output stream and then spin down the process. Please check with Luke on the correct approach for his transform process and transform functions.
1. For replace, I think there may be scenarios where a user would want to stop the data process, possibly edit the params or switch the input stream then relaunch. Again, I would check with Luke to see the best practice here for his transform framework.
2. We also need better testing of the transform tables. There is a test that demonstrates this capability but please verify that it is current, do a little performance testing as well as try different scenarios of how the table is attached and passed into the transform.
3. Chris W has been making progress on the data processes that are required for 2.1/ We will need to see what variations are required but you may want to experiment with some of the patterns that Michael has outlinedhttps://docs.google.com/drawings/d/1yhLJetOinVTrLEztqOe_w3lKXE_kJKSmArqHibDu29I/edit, primarily the HA pattern and multi-input if they are ready.

I think we may run into a few issues as we dig into remove and replace logic. We can talk thru the orchestration steps but first see what works and what does not.

While creating or updating tests, think about how the services could be improved to facilitate these operations and reduce the effort for a marine operator to manage his assets. Also the tests should verify that the expected events are sent.

Please let me know about any questions or ideas to improve the service operations. Thanks.

Maurice

Maurice Manning
858.822.4433
mmmanning@ucsd.edumailto:mmmanning@ucsd.edu

PR #223, opened 2012-08-07
Modified the publish event at a given time method of UNS to fit in with the new interface of the scheduler

swarbhanu:
Modified the publish event at a given time method of UNS to fit in with the new interface of the scheduler service. Fixed test for the same. This test relies on the fact that the scheduler service is included in r2deploy.yml, which is included in a new pull request pending in ion-definitions.


swarbhanu:
This pull request depends on changed ion defs which are in PR 188 of ion-definitions. The latter PR needs to be merged before this one can by automatically merged.

PR #227, opened 2012-08-08
To clean buildbot: Put skips on tests that are not working until they can be resolved.

swarbhanu:


swarbhanu:
Pull request: 228 should be merged first. That will allow the test, test_ctd_transforms_no_sim.py, to run fine with no exceptions.

PR #229, opened 2012-08-10
Resolves nostests error in test_workflow.py caused by previously different naming of parameter contexts for conductivity etc in different modules 

swarbhanu:
Uses a common way of creating a parameter_dictionary through CoverageCraft in all the ctd transform modules. This ensures that the different modules do not use different names when creating or reading parameters leading to no errors due to that. Also, some other bugs were fixed.

### DIFF
--- a/ion/processes/data/transforms/ctd/ctd_L0_all.py
+++ b/ion/processes/data/transforms/ctd/ctd_L0_all.py
@@ -27,6 +27,7 @@ from prototype.sci_data.stream_parser import PointSupplementStreamParser
 from ion.services.dm.utility.granule.record_dictionary import RecordDictionaryTool
 from ion.services.dm.utility.granule.taxonomy import TaxyTool
 from ion.services.dm.utility.granule.granule import build_granule
+from ion.services.dm.utility.granule_utils import CoverageCraft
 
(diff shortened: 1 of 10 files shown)


MauriceManning:
So CoverageCraft is basically just for CTD granules? we should make tat explicit so that AS and other do not think this will  allow them to create their data types or explain how it can be extended. Im a little worried about this convenience  method getting used too pervasively.

swarbhanu:
Maurice, the CoverageCraft methods are just a convenient simple way to make parameter dictionaries. If in any AS test, one needs a parameter dictionary that will contain special variables like 'google_dt' then one can follow the example pattern in CoverageCraft.create_parameter() and do it manually. Or one can just include that variable inside the common CoverageCraft method. As far as I understand there is no harm done in using a parameter dictionary in tests which holds the contexts for more variables than one actually needs in the test. It is fine to write them separately in code without using a common method, but that will lead to (what happened actually... reason why tests were failing) that one developer uses a name 'pres' and another in another module uses the name 'pressure', and then if the tests use modules in a complicated way, then there will breaking of tests. So I think CoverageCraft should be regarded as just a convenient way of creating things like parameter dictionaries which everyone uses. If it needs to be changed, one can just change it there in that one module and all tests using it will still pass.

PR #230, opened 2012-08-10
Replaced "while true" with the pattern, "while not self.finished.is_set()" in ctd data producer modules

swarbhanu:
Replaced the while True with Dave's pattern, while not self.finished.is_set(), so that the loop ends when the process quits in ctd data producer modules.


swarbhanu:
This PR is based on top of PR 229. So it might be more proper to merge PR 229 first and then this one. I kept the PRs separate because they were dealing with different issues.

swarbhanu:
This PR references the PR 190 in ion-definitions. Therefore the ion-def PR should be merged first before this one.

swarbhanu:
The git checkout fail in the buildbot happened because the ion-defs were not merged. But I have merged it myself now since Raj also confirmed that was what he wanted.

swarbhanu:
I will make another pull request and that should work.

PR #231, opened 2012-08-10
This should get buildbot running with a correct git checkout (unlike PR 230)

swarbhanu:


swarbhanu:
This PR is the same as PR 230 which failed due to a git checkout error. Merged this myself as it does not change any code at all relative to PR 230.

PR #237, opened 2012-08-16
The scheduler integrated with the processing of batch notifications

swarbhanu:
Tests complete and working for batch processing using the scheduler. The operator will now use a method, set_process_batch_key(), to set up the process_batch_key parameter that the UNS will listen for as origin of the timer event published by the scheduler which will kickstart the batch processing of notifications. Bumped ion defs to make that possible.


swarbhanu:
The ion defs bumped are in master of ooici/ion-definition. So one can just merge this pull request.

PR #244, opened 2012-08-22
Yaml for parameter dicts and parameter context definitions

swarbhanu:


swarbhanu:
This PR is based on top of PR 243. It is only the last commit that is relevant to the adding of the yml files for param dicts defs and param context defs. The other commits are actually PR 243.

One can merge the two PRs, PR 243 and 244, in any order without any problems.

PR #265, opened 2012-08-30
Fixes test in which notification worker sends notification email. This fixes buildbot.

swarbhanu:
Now that notification workers are being created on bootstrap the way ingestion workers are, an error was appearing in an integration test for notification workers because the test was creating its own worker and only checking whether that worker had sent an email. In reality, however, one of the other workers in the container was doing so. This patch in the test finds the worker processes in the container and checks if an email has been sent. As before, the test also parses the message body and checks that it has the right information and also makes assertions about the recipient and sender.


daf:
Great, thanks!

PR #270, opened 2012-09-05
Added binary param context for binary data. Bumped ion-def with the new context added

swarbhanu:


ion/util/parameter_yaml_IO.py
```diff
@@ -75,6 +75,12 @@ def build_contexts():
     depth_ctxt.fill_value = 0e0
     contexts.append(depth_ctxt)
 
+    binary_ctxt = ParameterContext('binary', param_type=QuantityType(value_encoding=np.int64))
```
lukecampbell, line 78: I don't know if this is a good fit, a 64bit integer is not representative of a 'byte'.

blazetopher:
If this can be tabled for a little longer (like the rest of today) - we should have the appropriate handling for this datatype, which is not using the QuantityType but the ArrayType (to facilitate byte arrays of varying length per record)

swarbhanu:
Yes, sounds good. I was not actually sure of what was going to be included as 'binary' data. Yes, Chris would have a better idea of that. Thanks.

lukecampbell:
OBE

PR #275, opened 2012-09-13
Removed coverage craft from sa tests

swarbhanu:
Using the get_param_dict() method to pull param dict from yml for now. Soon we will add the feature where it gets param dict from the couch.

This pull request only affects SA test code and so it should not break anyone's code.


lukecampbell:
Merged into `refactor_pubsub`

PR #283, opened 2012-09-19
await_process_state function for thread-safe process status checks

ianfixes:
this will need a corresponding change to ion-definitions if approved.

### DIFF
--- a/ion/services/cei/process_dispatcher_service.py
+++ b/ion/services/cei/process_dispatcher_service.py
@@ -13,6 +13,8 @@ from pyon.core.exception import NotFound, BadRequest, ServerError
 from pyon.util.containers import create_valid_identifier
 from pyon.event.event import EventPublisher
 from pyon.core import bootstrap
+from pyon.event.event import EventSubscriber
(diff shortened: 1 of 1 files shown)


ianfixes:
@shenrie @mmeisinger 

David L. asked me to have you approve this

mmeisinger:
I like the idea of the ProcessStateGate. I don't like to add the function await_process_state as a service operation to PD.

A process waiting for another process to come up should import the ProcessStateGate and then use it directly as an event subscriber. There is no need to use a PD operation for this, especially a blocking RPC. Remember that every RPC blocks the worker process (here: the PD) for the time until return, is subject to an RPC timeout and has a messaging overhead and latency.

So yay on the ProcessEventGate (and even extension helpers as needed), yay to place it into the process_dispatcher.py for now unless David L has a better place for process related utils (maybe even in pyon?) and nay on the PD service operation

ianfixes:
I thought that might be the case.  We could put ProcessStateGate in pyon (pyon/event/event.py), which is where EventGate lives.  Would that make more sense?

PR #289, opened 2012-09-28
Luke: NoneType problem with ProcessStateGate

lukecampbell:
If in_desired_state is called before a process is scheduled, the
read_process will return None.  When the method is called it will raise
a AttributeError, Nonetype has no attribute process_state.

This catches this case and simply returns False


ianfixes:
return process_obj and self.desired_state == process_obj.process_state

lukecampbell:
I will take that action

PR #294, opened 2012-10-03
Adds assertions to test for ctd_L0_all transform. Also adds tests for L1 pressure and temperature transforms

swarbhanu:
Checks that the transform algorithms are successfully executed.


swarbhanu:
Int tests that verify the execution of algorithms for all the L0, L1 and L2 transforms have been completed.

PR #315, opened 2012-10-19
Depth to pressure

swarbhanu:
Converted SA int tests to use pressure and removed 'depth'. Made the same changes in other transform code and in simple_ctd_publisher and other example transforms and publishers where granules are made.


swarbhanu:
One more fix is needed. Temporarily closing the PR. It will be opened again very soon.

PR #321, opened 2012-10-23
Removes transforma. Fixes on_quit() of a couple of transforms by calling the base class.

swarbhanu:


ion/processes/data/simple_granule_logger.py
```diff
@@ -20,6 +20,7 @@ def on_start(self):
         self.log       = self.CFG.get_safe('process.log', False)
 
     def on_quit(self):
+        super(SimpleGranuleLogger, self).on_quit()
         self.subscriber.stop()
```
daf, line 24: This line is called by `on_quit` already, so calling `stop` again is likely going to error out.  Also, this class doesn't call `on_start`'s base method.  Luke was saying that some of these are intentionally not calling their baseclass, this might be one of them.

swarbhanu, line 24: Okay, thanks for pointing it out. I will revert it now.

Thanks,
Swarbhanu.

ion/services/sa/test/test_activate_instrument.py
```diff
@@ -344,7 +344,7 @@ def test_activateInstrumentSample(self):
         self.addCleanup(self.imsclient.stop_instrument_agent_instance,
                         instrument_agent_instance_id=instAgentInstance_id)
 
-        gevent.sleep(2)
+        gevent.sleep(4)
```
lukecampbell, line 347: You should never be using a `gevent.sleep` try to make it either asynchronous using gevent events or use a polling mechanism.

ion/processes/data/simple_granule_logger.py
lukecampbell, line 24: You can leave it, this class isn't really used right now.

swarbhanu, line 24: Right. Reverting it now.

ion/services/sa/test/test_activate_instrument.py
swarbhanu, line 347: Yes, I will change that soon. Should be by today evening. Thanks.

PR #324, opened 2012-10-23
Observatory management's create_site_data_product() method updated

swarbhanu:
The create_site_data_product() method of observatory management service has been updated so that it creates a data process and launches it instead of using cei to create a process definition and launch it through a process_dispatcher.schedule_process().

No one should be affected by the code change as it only affects the above method of observatory management in the described way.

However, to make things work, the create_data_process() method in data_process_management was made stable for the case where one uses it but does not pass in any input data product ids (which is the usual situation in observatory related stuff). The method was made stable so that it will work correctly even if no input data product ids are provided.

Also, the logical transform was converted to the new transform design.

### DIFF
--- a/ion/processes/data/transforms/logical_transform.py
+++ b/ion/processes/data/transforms/logical_transform.py
@@ -1,21 +1,28 @@
-import commands
-import time
 from gevent.greenlet import Greenlet
-from interface.services.cei.iprocess_dispatcher_service import ProcessDispatcherServiceClient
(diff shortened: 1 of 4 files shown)


shenrie:
I don't understand why you would not want to use the CEI process dispatcher.

swarbhanu:
Stephen, Maurice suggested that we create data_process and launch it because that will be similar to the way we are doing similar stuff elsewhere.

PR #332, opened 2012-10-26
Luke: patches a tests assertion, still needs work

lukecampbell:


ion/services/ans/test/test_workflow.py
```diff
@@ -339,10 +340,15 @@ def test_mpl_graphs_transform_workflow(self):
...
-        #Validate the data from each of the messages along the way
-        self.validate_mpl_graphs_transform_results(retrieve_granule)
+#        def poller():
+#
+#            retrieve_granule = self.data_retriever.retrieve(ds_ids[0])
+#
+#            #Validate the data from each of the messages along the way
+#            return self.validate_mpl_graphs_transform_results(retrieve_granule)
+#
+#        poll(poller, timeout=10)
```
lukecampbell, line 351: DM is an asynchronous framework; clients cannot rely on retriever to give you an entire dataset before ingestion has ingested the dataset.
These lines are commented out because ingestion does not actually ingest the mpl dataset.  http://cl.ly/image/1q0q1d282f2b

Also, the MPL param dict def reads as follows:

```
mpl_graph: ['mpl_graph'] # will never go into a coverage, do not try to ingest, it will die, painfully.
```

If this data needs to be ingested, the parameter dictionary definition will need to be altered: _by adding a `time` parameter_

ion/services/ans/test/test_helper.py
```diff
@@ -438,6 +441,8 @@ def validate_mpl_graphs_transform_results(self, results):
                     assertions(graph['viz_product_type'] == 'matplotlib_graphs' )
                     # check to see if the list (numpy array) contains actual images
                     assertions(imghdr.what(graph['image_name'], h = graph['image_obj']) == 'png')
+                    found_data = True
+        return found_data
```
lukecampbell, line 440: Preps the `validate_mpl_graphs_transform_results` method to be used in a polling pattern.

ion/services/ans/test/test_helper.py
```diff
@@ -163,6 +163,8 @@ def message_received(message, stream_route, stream_id):
                 result.set(True)
 
         subscriber = StandaloneStreamSubscriber(exchange_name='workflow_test', callback=message_received)
+        subscriber.xn.purge()
+        self.addCleanup(subscriber.xn.delete)
```
lukecampbell, line 167: adds proper cleanup to the queues.

PR #345, opened 2012-11-06
Exchange Bootstrap plugin for on_init

daf:
Exchange bootstrap plugin expansion.

Creates exchange spaces and exchange points as defined in pyon.yml.  Also creates brokers and associates them with the exchange space
- Also adds idempotency to create methods in EMS and fixes bugs in XS delete.


mmeisinger:
Looks right on

PR #356, opened 2012-11-13
Luke: fixes the startkey/endkey issues with bigcouch

lukecampbell:


mmeisinger:
Eventually (in transition if possible), all queries to the datastore should be moved to the find_by_view function. We currently use 2 different, overlapping functions doing essentially the same (flexibly query an index with arguments).

PR #378, opened 2012-11-27
Bump pyon to include rr client indirection, fix access in OMS

daf:


daf:
Several services changes their access to directly use `self.container.resource_registry`, I will address these later this afternoon.

PR #385, opened 2012-12-04
Fixes UNS for end to end demo using ION-UX

swarbhanu:
Patches uns_utility_method.py so that it can calculate the subscribed users correctly when notification requests are made with event_subtype set as ''. Made the method more resilient against different kinds of inputs.

Added the option to pass an origin to the demo stream alert transform when it is launched by sending it a  CFG.process.event_origin field. But this is just an option and it does NOT have to be passed the origin. The existing preload code will work fine. Existing tests which do not pass the origin will work fine too. When the origin is not being passed (as in several successfully running tests), the transform extracts the origin from the granule using its data_producer_id attribute.

The code now has been verified to atleast send fake emails using a fake smtp client to the correct user email address using the correct sender's address and with the right contents.

What is left is to try it from inside the UCSD VPN to send a real email using mail.oceanobservatories.org.

### DIFF
--- a/ion/processes/data/transforms/event_alert_transform.py
+++ b/ion/processes/data/transforms/event_alert_transform.py
@@ -138,7 +138,7 @@ class DemoStreamAlertTransform(TransformStreamListener, TransformEventListener, 
         self.timer_interval = None
         self.count = 0
         self.timer_cleanup = (None, None)
-        self.origin = "Not yet computed as no earlier granule was received from instrument"
+        self.origin = ""
 
     def on_start(self):
         super(DemoStreamAlertTransform,self).on_start()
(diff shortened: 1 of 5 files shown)


lukecampbell:
DO NOT MERGE!

extern/pyon
```diff
@@ -1 +1 @@
-Subproject commit e158fa325f20d769f4536338e204d6482365e93b
+Subproject commit d7747f5df27d5f8ef4e072d55c445e369d91199b
```
lukecampbell: This is reverting our changes twice now, please do not merge this.

extern/coverage-model
```diff
@@ -1 +1 @@
-Subproject commit 4138c089624c665dc7fab7b2db391feb4f5678ff
+Subproject commit ab9e3e201c524778927e16d982513501be003553
```
lukecampbell: This is reverting our changes, please do not merge this.

swarbhanu: Yes. Not merging this. Will fix this.

daf:
Looks good now.

lukecampbell:
Looks good, ready to merge.

PR #388, opened 2012-12-05
update to driver launcher to support running driver from egg

unwin:


wfrench:
This is only a portion of the work that needed to be done in coi services to enable egg launching.  We will close this request from now and merge all at once.

PR #389, opened 2012-12-06
Update to driver_process module to add driver launching via an egg

wfrench:
This update puts the changes in place to start launching drivers from eggs.  However, changes still need to be made to IMS or where ever the driver_config dictionary is created to start using this feature.

Also, it appears that the instrument agent no longer uses the packet factory class.  So I've removed that code from the instrument agent and driver process module.


edwardhunter:
Looks good.

PR #391, opened 2012-12-06
UNS is now working fine with UX

swarbhanu:
No external api change was made to UNS.

The smtp client is created immediately before sending an email to avoid timeout issues when there is no activity.

The code was cleaned up so that unnecessary legacy code was cleaned up.

All tests are working.

UNS can now be used in the UX system and it has been verified that notification emails do get sent after subscriptions are created and the platform agent is in running mode.

Added description to the notification emails.


daf:
Looks good, thanks.

One thing to note is that we'd like to remove the `self.smtp_client` in the future as we don't need to store it as an instance attribute but Swarbhanu tells me this would require a lot of test correction.  It's safe now due to the single-threaded style execution of the callback from the endpoint/subscriber/transform code.  Just something that would be nice to change to make the code more concise later on.

PR #404, opened 2012-12-11
(fix) Broken preload

newbrough:
there was an issue with pubsub association call (missing predicate) which broke ability to read streamdef.

after fixing issue, can now merge in maurice's updates to preload.

### DIFF
--- a/ion/services/dm/distribution/pubsub_management_service.py
+++ b/ion/services/dm/distribution/pubsub_management_service.py
@@ -42,7 +42,9 @@ class PubsubManagementService(BasePubsubManagementService):
                 return existing[0]
             raise Conflict('StreamDefinition with the specified name already exists. (%s)' % name)
 
-        if not name: create_unique_identifier()
+        if not name:
+            create_unique_identifier()
+#            name = create_unique_identifier()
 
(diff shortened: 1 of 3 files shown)


edwardhunter:
Reviewed.

ion/services/dm/distribution/pubsub_management_service.py
```diff
@@ -42,7 +42,9 @@ def create_stream_definition(self, name='', parameter_dictionary=None, parameter
                 return existing[0]
             raise Conflict('StreamDefinition with the specified name already exists. (%s)' % name)
 
-        if not name: create_unique_identifier()
+        if not name:
+            create_unique_identifier()
```
daf, line 46: This is a no-op.  Shouldn't it be the commented out line?  Otherwise, why call it at all?

ion/services/dm/distribution/pubsub_management_service.py
```diff
@@ -42,7 +42,9 @@ def create_stream_definition(self, name='', parameter_dictionary=None, parameter
...
+#            name = create_unique_identifier()
```
daf, line 47: This line might be the correct one?

lukecampbell, line 47: it really should just be `name = name or create_unique_identifier()` I'll patch it in a bit.

PR #406, opened 2012-12-12
Adds buffering to ingestion workers, decreases overall memory use.

lukecampbell:
Addresses memory concerns.

Part of [OOIION-580](https://jira.oceanobservatories.org/tasks/browse/OOIION-580)
## Memory Loss in Ingestion

There is a concerning memory leak inside the ingestion layer. Currently, ingestion workers can only consume messages 
for a few hours before running the system dry of memory. There are a few areas where the leak is stemming from:
- h5py has a small leak when it comes to opening and closing files, detailed in [issue 248](http://code.google.com/p/h5py/issues/detail?id=248)
- h5py also has a larger leak when it uses variable length strings as a data type, I patched this with h5py's author [here](https://github.com/lukecampbell/h5py/compare/master...patch1)
- our messaging layer also consumes memory
### Patching `h5py`

We coordinated with the lead developer on `h5py` and we patched the variable length string consumption leak. The 
patch is not an official release so it's being referenced from my github repository until it is released as an official
patch. 
### Patching Ingestion

One of the big problems with ingestion and the coverage model is how many times it opens an HDF file, which still 
consumes a small amount of memory. The coverage model opens an HDF file once for each parameter for each insertion (regardless of size), so in a coverage with
ten parameters, it will open at least ten files for each granule which can add up over time. After understanding this
better, we added an internal buffer into the ingestion workers. The buffer holds a specified amount of granules, once
the buffer fills up they are flushed to coverage as one, this reduced the memory consumption by almost half.
### Messaging

There is still some memory consumption (since it's pure python it might not be a leak) in the messaging layer. At a
cursory profile it looks like there are hanging references to ConversationEndpoints and dictionaries. After all the
other parts were patched, messaging became the majority consumer of memory. Messages compose 93% of the ingestion
process' memory consumption. One way to wrap up buffering and limit the messages is to utilize our bandwidth better by
buffering messages closer to the instruments which limits the traffic and maximizes our bandwidth efficiency.
## Consumption Rates Broken Down

<table>
<tr><th>mode</th><th>size / message</th><th>Consumption/h (MB)</th></tr>
 <tr><td>Full Ingestion</td><td>9.8</td><td>34.453</td></tr>
 <tr><td>No Ingestion</td><td>9.22</td><td>32.414</td></tr>
 <tr><td>Raw Messaging</td><td>6.493</td><td>22.6</td></tr></table>

- _Full Ingestion_ is the full ingestion layer (after the h5py patch and buffering)
- _No Ingestion_ still launches an ingestion worker but the worker simply ignores all messages
- _Raw Messaging_ is just a subscriber and publisher streaming messages every second

The sampling was done by measuring **four** producers publishing granules of length 1 to four consumers (ingestion
workers).


blazetopher:
We will need to revisit the RDT.append() function as it may not suffice as we start handling more complex data streams - but it will do the job for R2.0.  The proper way to "buffer" samples is at the producer end - this is where the functionality should ultimately end up if possible.

Looks good.

PR #407, opened 2012-12-12
Prevents registration from clobbering during restart.

lukecampbell:
This hopefully will address [OOIION-607](https://jira.oceanobservatories.org/tasks/browse/OOIION-607)


blazetopher:
Seems a good thing to check! :)

PR #411, opened 2012-12-13
Updated test code for code coverage 

seman:
Updated test code for code coverage

### DIFF
--- a/ion/services/coi/test/test_conversation_management_service.py
+++ b/ion/services/coi/test/test_conversation_management_service.py
@@ -152,6 +152,12 @@ class TestPolicyManagementService(PyonTestCase):
         self.assertEqual(ex.message, "Conversation Role 'bad role' does not exist")
         self.mock_read.assert_called_once_with('bad role', '')
 
+    def test_not_implemented_methods(self):
+        self.convo_management_service.create_conversation()
+        self.convo_management_service.update_conversation()
+        self.convo_management_service.read_conversation()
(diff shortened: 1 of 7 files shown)


ion/services/coi/test/test_conversation_management_service.py
```diff
@@ -181,10 +187,17 @@ def setUp(self):
...
+        bad_obj = IonObject('ConversationType', name='bad name', definition=self.rpc_scribble_protocol)
+        with self.assertRaises(BadRequest):
+            self.convo_management_service.create_conversation_type(bad_obj)
+
         convo_type_obj = IonObject('ConversationType', name='RPC', definition=self.rpc_scribble_protocol)
         convo_type_id = self.convo_management_service.create_conversation_type(convo_type_obj)
         self.assertNotEqual(convo_type_id, None)
 
+        with self.assertRaises(BadRequest):
+            self.convo_management_service.read_conversation_type()
```
daf, line 199: You aren't required to use the context-manager style call with `assertRaises` - typically you'd use that if you wanted to test something about the exception itself, like the contents of the message.  You could rewrite this test as (notice the lack of `()`):

``` python
self.assertRaises(BadRequest, self.convo_management_service.read_conversation_type)
```

This is minor and I wouldn't expect you to go back and change all these, just for the future.

daf:
Looks like this just updates test coverage, let's get it in there.

PR #412, opened 2012-12-13
Ingestion

lukecampbell:
- Improves the buffer management in ingestion workers.
- Adds a timeout before flushing the queues
- Adds configurations to configure the time limit and the buffer limits.
- Adds a default value of 1 in `pyon.yml` so that tests pass
# NOTE: ON PRODUCTION SYSTEMS YOU MUST SPECIFY AN ADEQUATE BUFFER AND TIME LIMIT IN `pyon.local.yml`

### DIFF
--- a/README_DEMO
+++ b/README_DEMO
@@ -21,8 +21,17 @@ Prepare for start:
 Start the services container:
 > bin/pycc -fc -r res/deploy/r2deploy.yml
 
+Preload system resources (without OOI resources or UI)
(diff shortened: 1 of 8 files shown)


blazetopher:
Reviewed - this looks like a good 'backup' strategy for when the source of the published data either can't or won't buffer data samples.  Enables ingestion to combine multiple granules to reduce the number of times the coverage is called for write operations.

The preferred workflow is still that the publisher buffers such that more than one sample is in each granule.

blazetopher:
Final review - merged.

PR #414, opened 2012-12-13
Shoud fix the status roll up seen through the UI (OOIION 622)

swarbhanu:
Fixes the timestamp being published for the DeviceCommsEvent so that it uses NTP time.

Have verified that the status icon (which resembles the power symbol) changes to good/bad when the UI page is refreshed.

This, therefore, should fix the status roll up.

All tests working.


MauriceManning:
this is a quick fix. time_stamps are not used in events, use ts_created. To be cleaned up correctly asap.

PR #415, opened 2012-12-14
fix preloader messages, restore working unit test

newbrough:
improved debug and error messages in preload

fixed nosetest to only use NOSE scenario so it won't break each time names of resources are changed for demos.


mmeisinger:
Looks OK

PR #416, opened 2012-12-14
Event Persister listens on a named queue, fixes OOIION-620

daf:


daf:
There needs to be a suitable "teardown" for system shutdown that removes this created queue.  Do we have such an "anti-bootstrap" stage?

daf:
(the above does not apply to restart - the purpose of naming the queue is to ensure restart)

mmeisinger:
Very good

PR #418, opened 2012-12-17
Changes loop conditions to use gevent Events in ingestion

lukecampbell:
Ingestion's internal timer was using a bi-conditional while-loop to
flush the queues, which caused unnecessary latency in on_quit. This
patch converts the loop to use gevent.event.Event which exits
immediately.


daf:
:+1:

PR #420, opened 2012-12-17
Adds support for inline data writes.

lukecampbell:


blazetopher:
Reviewed.  Out-of-band data writing via zmq can be used by setting 'service.ingestion_management.inline_data_writes' == False.  NOTE: this flag applies for the LIFETIME of the coverage!!  It cannot be changed from one load to another.

PR #424, opened 2012-12-19
Part of OOIION-619 and fixes buildbot (test_demo_stream_granules_processing) 

swarbhanu:
- Two event subscribers that the UNS and notification worker creates now use named queues.
- The test_demo_stream_granules_processing() method in test_transform_prototype.py now skips CEI launch. This was needed because of an additional test piece that was added recently to it. This test piece involves using the container's datastore to find events stored in the events db after they get created during the test run. The test works well in a single container and is very useful as an additional means to verify that the stream alert transform works. But in a multi-container test, the test does not work because of the above reason. The test should therefore skip CEI launch.

### DIFF
--- a/ion/services/dm/presentation/user_notification_service.py
+++ b/ion/services/dm/presentation/user_notification_service.py
@@ -204,6 +204,7 @@ class UserNotificationService(BaseUserNotificationService):
         """
         self.batch_processing_subscriber = EventSubscriber(
             event_type="ResourceEvent",
+            queue_name='user_notification',
             origin=process_batch_key,
             callback=process
         )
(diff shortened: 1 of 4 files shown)


ion/services/dm/presentation/user_notification_service.py
```diff
@@ -204,6 +204,7 @@ def process(event_msg, headers):
         """
         self.batch_processing_subscriber = EventSubscriber(
             event_type="ResourceEvent",
+            queue_name='user_notification',
```
daf, line 207: This is fine as long as UNS remains a singleton.  For the future, if we want to expand the number of UNS instances, having a named queue here is fine, but the callback method will should extract the process_batch_key out of the headers (or make it a lambda call, passing in the current value of it), that way round-robin delivery to all UNS instances would be safer.

There would need to be an additional step where the binding is removed from the queue when we want to tear down this batch processing.

Again, for the future.  Comments for posterity.

swarbhanu, line 207: Cool. Adding a R2.1 task for this.

PR #425, opened 2012-12-19
Remove assert from UNS set_process_batch_key handler

daf:
Using a named queue in this EventSubscriber, multiple
UNS instances will receive these messages in round-robin.
They will likely not line up with the exact instance that
created the subscriber with this exact binding/batch_key.

This will at least allow them to proceed.


daf:
(other work surely needs to be done to allow multiple UNSes to work properly)

PR #427, opened 2012-12-19
Schedule service on_quit cancels pending scheduled events

daf:
Currently, `on_quit` does not attempt to handle the greenlets started with `spawn_later` that is the core of the SchedulerService.  This patch attempts to fix that, with respect to the possible coming restart.

I notice there is an `on_system_restart` but I am not sure where this is called from - maybe a bootstrap service call?  That would make the most sense.

This PR for review please.

### DIFF
--- a/ion/services/cei/scheduler_service.py
+++ b/ion/services/cei/scheduler_service.py
@@ -17,7 +17,12 @@ import gevent
 
 
 class SchedulerService(BaseSchedulerService):
-    schedule_entries = {}
+
+    def __init__(self, *args, **kwargs):
+        BaseSchedulerService.__init__(self, *args, **kwargs)
+
+        self.schedule_entries = {}
+        self._no_reschedule = False
 
     def on_start(self):
         if CFG.get_safe("process.start_mode") == "RESTART":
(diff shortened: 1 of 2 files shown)


seman:
I am not sure if line [41](https://github.com/daf/coi-services/blob/2a89957992e68a9dd00dde891cdba30870774c41/ion/services/cei/scheduler_service.py#L41) will work.  We need to get the keys rather than the index. Also when calling `__delete()`, set the `force` delete to True. Otherwise, it will only delete expired timers. 

You right, `on_system_restart` is called from bootstrap service to restore timers during restart.

daf:
You may be looking at the first commit only - I removed the call to `__delete` as that was incorrect (didn't want to delete from RR).

That also probably means I don't need the `enumerate` call because I'm not using `idx` anywhere, will push a cleanup in a second.

seman:
You are right, we don't want to remove it from RR. There is a similar code in line [229](https://github.com/daf/coi-services/blob/2a89957992e68a9dd00dde891cdba30870774c41/ion/services/cei/scheduler_service.py#L229-234) to do this.

daf:
Good point.  I'll factor these out into a common method.

mmeisinger:
Is this PR still in development or ready to be merged? Otherwise please close.

daf:
I think the follow-up commit made this correct.  Will merge.

PR #430, opened 2012-12-20
Added cleanup for test using a transform that was not stopping at the end of a test

swarbhanu:
Added clean ups to the test module, event_management_test.py, using addCleanup()


jamie-cyber1:
Looks good.

PR #435, opened 2012-12-26
Remove DeviceCommsEvent

swarbhanu:
The transform that monitors the platform agent will only use DeviceStatusEvent. Both OK and OUT_OF_RANGE status will be sent using the latter event.

DeviceCommsEvent is removed from the system.


mmeisinger:
What's the status of this PR?

swarbhanu:
I am closing the PR for now to be opened later if needed. In the last week or so, there has been several code pushed in concerning fetching communication status which I think is also used in the UI for status rollups. Though a couple of weeks ago, Maurice and Tim wanted DeviceCommsEvent to be removed, I think we need more discussion on this as a lot of communication status related stuff that has been recently added will now break. Hence closing this pull request for now until further discussion.

PR #437, opened 2012-12-28
Notification preferences (digest, realtime) - Implements part of OOIION-563

swarbhanu:
# What is implemented?

A user is now able to specify the manner in which notifications were sent. 

UNS delivery code includes logic to check this flag and send real time to only users who have requested that route. During batch processing, digests notifications are sent to only users that have that option.

The notification preferences include a flag for delivery_enabled (the default being true). If set to False, the user will not receive any notification emails, digest or real time.
# How notification preferences are set?

An example:

```
        from interface.objects import  NotificationPreferences, NotificationDeliveryModeEnum  

        notification_preferences = NotificationPreferences()
        notification_preferences.delivery_mode = NotificationDeliveryModeEnum.REALTIME
        notification_preferences.delivery_enabled = True

        user = UserInfo()
        user.name = 'user_1'
        user.contact.email = 'user_1@gmail.com'
        user.variables.append({'name' : 'notification_preferences', 'value' : notification_preferences})

```
# What if the notification preferences are not set at all for a user?

No harm done. This pull request is backward compatible!

The code checks to see if the notification preferences have been set. If not, then it goes along the way it used to earlier. So if you have not set notification preferences for your user, there will be no harm done.  Nothing should break.
# What is not implemented?

We have not yet implemented logic that limits number of emails sent to a user in a day,... or hour?... since more clarification is needed regarding the appropriate time window. Also, more clarification is needed regarding whether the UNS will coordinate with a interval timer (scheduler service) in order to determine whether the above limit has been reached.

### DIFF
--- a/ion/processes/data/transforms/notification_worker.py
+++ b/ion/processes/data/transforms/notification_worker.py
@@ -143,10 +143,14 @@ class NotificationWorker(TransformEventListener):
 
         for user in users:
             notifications = []
+            notification_preferences = None
             for variable in user.variables:
                 if variable['name'] == 'notifications':
                     notifications = variable['value']
 
-            user_info[user._id] = { 'user_contact' : user.contact, 'notifications' : notifications}
+                if variable['name'] == 'notification_preferences':
+                    notification_preferences = variable['value']
+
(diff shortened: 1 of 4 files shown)


daf:
Looks good.

PR #438, opened 2012-12-31
Makes UNS more efficient by refactoring UNS methods in order to use the datastore once for a list of ids instead of multiple times

swarbhanu:
Uses read_mult() wherever possible instead of read().

The UNS API is not changed at all by this pull request.

### DIFF
--- a/ion/services/dm/presentation/user_notification_service.py
+++ b/ion/services/dm/presentation/user_notification_service.py
@@ -227,7 +227,7 @@ class UserNotificationService(BaseUserNotificationService):
         if not user_id:
             raise BadRequest("User id not provided.")
 
-        log.debug("Create notification called for user_id: %s, and notification: %s" % (user_id, notification))
+        log.debug("Create notification called for user_id: %s, and notification: %s", user_id, notification)
 
         #---------------------------------------------------------------------------------------------------
         # Persist Notification object as a resource if it has already not been persisted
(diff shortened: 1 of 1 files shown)


ion/services/dm/presentation/user_notification_service.py
```diff
@@ -699,12 +694,9 @@ def process_batch(self, start_time = '', end_time = ''):
                 # get the list of ids corresponding to the events
                 ret_vals = self.discovery.parse(search_string)
 
-                for event_id in ret_vals:
-                    datastore = self.container.datastore_manager.get_datastore('events')
-                    event_obj = datastore.read(event_id)
-                    events_for_message.append(event_obj)
+                events_for_message += self.datastore.read_mult(ret_vals)
```
daf: Nitpick, use `extend` instead of `+=`.

ion/services/dm/presentation/user_notification_service.py
```diff
@@ -868,21 +860,28 @@ def get_subscriptions(self, resource_id='', include_nonactive=False):
...
 
         notifications_all = set()
         notifications_active = set()
 
+        object_ids = []
         for item in ret_vals:
-
             if item['_type'] == 'NotificationRequest':
-                notif = self.clients.resource_registry.read(item['_id'])
+                object_ids.append(item['_id'])
```
daf, line 871: For a block like this, consider the use of a list comprension:

```
object_ids = [item['_id'] for item in ret_vals if item['_type'] == "NotificationRequest"]
```

Obviously they are not a panacea, they can get difficult to read with any more complexity, but for a case like this, it can actually improve readability.  Don't bother updating this one but just a note for the future.

ion/services/dm/presentation/user_notification_service.py
swarbhanu: Cool. I pushed the change in: extend for lists instead of +=.

daf:
This looks great!

ion/services/dm/presentation/user_notification_service.py
swarbhanu, line 871: Yes, good point. I will try to make things more pythonic .

PR #440, opened 2013-01-02
Pydap handler test

tgiguere:
Adds a module to unit test pydap coverage handler


blazetopher:
Reviewed - initial set of tests for CoverageModel handler for PyDap.  Test code only.

PR #443, opened 2013-01-04
Remove dm datastore

lukecampbell:
These patches fix bugs associated with tests only, this was originally part of another feature but it was separated to maintain a good workflow.


blazetopher:
Reviewed.

PR #447, opened 2013-01-07
Fixes OOIION-590 (Fixes UNS regarding emails being sent to events not subscribed to)

swarbhanu:
The UNS utility method designed to figure out interested users was not properly handling the case of events coming from the same origin and whose event subtypes and origin_type fields were empty.

This pull request patches the method so that it properly handles all cases.

It has been verified that only emails for the event subscribed to is being sent now.

### DIFF
--- a/ion/services/dm/presentation/test/user_notification_test.py
+++ b/ion/services/dm/presentation/test/user_notification_test.py
@@ -902,23 +902,17 @@ class UserNotificationIntTest(IonIntegrationTestCase):
 
         notification_request_1 = NotificationRequest(   name = "notification_1",
             origin="instrument_1",
-            origin_type="type_1",
             event_type='ResourceLifecycleEvent',
-            event_subtype=''
         )
 
         notification_request_2 = NotificationRequest(   name = "notification_2",
             origin="instrument_2",
-            origin_type="type_2",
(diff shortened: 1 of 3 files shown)


daf:
Two comments on style (not related to content):

##### 1: use comma syntax on log statements, not % interpolation

For instance, you have several log statements like this:

```
log.debug("For event_origin_type = %s too, UNS got interested users here  %s" % (event.origin_type, set()))
```

Instead, use this form:

```
log.debug("For event_origin_type = %s too, UNS got interested users here  %s", event.origin_type, set())
```

The goal being that if the `DEBUG` level is not set, it doesn't have to build the string that it sends to the debug call.  The call will take care of interpolating the string for you.  (You also get to avoid the extra parentheses pair that % requires).

(also what is up with the `set()` as a log param, I don't think that's what you meant to have in that line).

##### 2: use `extend()` to append to a list, not `+=`

it's just more pythonic.

```
user_list_2.extend(reverse_user_info['event_origin'][''])
```

swarbhanu:
Yep, I will push the change in a moment.

swarbhanu:
Okay, so I pushed in the change in the commit below: 8cebc69. It addresses the better log debug style and the extend for lists.

PR #448, opened 2013-01-08
Mscweb

pkediyal:


daf:
Couple of comments:
- You'll need to merge master in your branch to resolve whatever Github is saying can't be auto-merged.  I'm assuming it's the ion-defs change, did you mean to change it?
- Since the html file is referencing d3 and jquery from CDNs, you likely don't need to include them in our codebase.

pkediyal:
ion-def changes must be synchronously pushed for conversation log to work. So far, Michael handles that mess and I am a bit unsure of my role in the protocol regarding how that is managed.

I did not change the html file nor its libraries. Nevertheless, I did incorrectly think that it was referencing local.

I will close this and make changes.

daf:
You don't have to close - push more commits to the same branch and they'll automatically show up here.  Keeps the flow going.

pkediyal:
so should I first create a pull request for ion-defs? then when that gets merged I can update the submodule in pyon

PR #452, opened 2013-01-09
Replay alignment

lukecampbell:
This patch handles a certain race condition where an ingested coverage is in the middle of extending the dataset when another coverage is instantiated during a retrieval request. The new coverage is not fully consistent and causes a mis-formed granule to be created.


blazetopher:
Reviewed.  Addresses OOIION-651.

blazetopher:
Reviewed - increase in default brick size to increase IO efficiency.

PR #457, opened 2013-01-10
Restore datastore service tests (despite not using this service)

daf:


mmeisinger:
Thanks, good solution

PR #462, opened 2013-01-11
Follow up to OOIION-619. Starting notification workers who know their queue_name.

swarbhanu:
The notification workers are meant to listen to events in round robin. Now, since pull request 423 has been merged, they can extend TransformEventListener as before and also process events in round robin. Also, they are listening on a named queue, which is good for debugging, incase something happens.

Since the queue of the parent transform class, TransformEventListener, is being used, the queue should get cleaned up properly when processes stop and the TransformEventListener's on_quit() gets called while quitting.


daf:
:accept:

PR #469, opened 2013-01-15
Adds conditional for closing coverage in retrieve_oob

lukecampbell:
There is a chance that _get_coverage could fail and if it does then
coverage is not defined when it reaches the finally block to close the
coverage, this patch addresses this case.

Addresses the failure [here](http://buildbot.oceanobservatories.org:8010/builders/r2_full/builds/260/)
Part of [OOIION-663](https://jira.oceanobservatories.org/tasks/browse/OOIION-663)


blazetopher:
Reviewed.  Addresses [OOIION-663](https://jira.oceanobservatories.org/tasks/browse/OOIION-663)

PR #470, opened 2013-01-15
Adds skipTest to oob_retrieve

lukecampbell:
Skips a test in the CEI environment, the test can't retrieve a coverage if the host does not share the filesystem where the coverages reside.


blazetopher:
Reviewed - test only

PR #471, opened 2013-01-15
Changes file modes in ingestion workers to a

lukecampbell:
Before it was using 'w' which wasn't respected, but going forward it
should be a to ensure the HDF file is not reset.

Part of [OOIION-660](https://jira.oceanobservatories.org/tasks/browse/OOIION-660)


blazetopher:
:+1:

PR #474, opened 2013-01-15
Adds short circuit None evaluation to rdt

lukecampbell:
If a parameter in an RDT needs to be deleted or removed for any specific
reason, this bypasses the checking done in the Coverage layer and just
sets the field to None.

Part of [OOIION-579](https://jira.oceanobservatories.org/tasks/browse/OOIION-579)


blazetopher:
Reviewed - looks fine to me.

blazetopher:
Reviewed - looks good

PR #482, opened 2013-01-16
Bumps coverage-model ref

blazetopher:
Makes 'r' the default mode when loading a coverage

Addresses [OOIION-668](https://jira.oceanobservatories.org/tasks/browse/OOIION-668)


lukecampbell:
:+1:

PR #488, opened 2013-01-21
Add /version route to ServiceGatewayService for UX

daf:


daf:
Please don't merge before review of UX changes.

PR #495, opened 2013-01-23
Adds replace data process functionality to data process management service

swarbhanu:
- Adds replace_data_process() to data process management to satisfy SA requirement for replace functionality.
- Adds test code to verify that that the replace method works.

For the replace functionality, what we are doing is:
1. Stop the running process (We do not yet have a pause functionality in process dispatcher). Its process_id is stored in the process_id attribute of the data process object
2. Update the data process object with the new params
3. Restart the process associated with the data process (and update the data process object's process_id with that of the new running process)

The id of the original data process object is maintained since the data process object is a resource object.


mmeisinger:
OBE

PR #498, opened 2013-01-24
Adds failover logic (and test) to ingestion workers

lukecampbell:
Ingestion workers now publish ExceptionEvents when corruption occurs,
the stream publishers, or a managing entity could listen for these
events and attempt to either clean the coverages or stop ingestion until
further investigation is complete.  In any case this is a step in
bullet-proofing the system.

Part of [OOIION-660](https://jira.oceanobservatories.org/tasks/browse/OOIION-660)

### DIFF
--- a/ion/processes/data/ingestion/science_granule_ingestion_worker.py
+++ b/ion/processes/data/ingestion/science_granule_ingestion_worker.py
@@ -7,6 +7,8 @@
 '''
 from ion.services.dm.utility.granule.record_dictionary import RecordDictionaryTool
 from interface.services.coi.iresource_registry_service import ResourceRegistryServiceClient
+from pyon.core.exception import CorruptionError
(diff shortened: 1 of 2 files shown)


ion/processes/data/ingestion/science_granule_ingestion_worker.py
```diff
@@ -43,7 +45,11 @@ def on_start(self): #pragma no cover
 
     def on_quit(self): #pragma no cover
         self.subscriber.stop()
-        self.flush_all()
+        try:
+            with gevent.Timeout(5):
+                self.flush_all()
+        except:
+            pass
```
lukecampbell, line 52: There was a bug where if there was an exception raised during this final flush attempt that the ingestion worker would just hang forever :unamused:. Now it get's 5 seconds to try to flush everything.

blazetopher:
:+1:

PR #501, opened 2013-01-24
Adds fuzzy striding to improve replay speeds.

lukecampbell:
Should solve [OOIION-645](https://jira.oceanobservatories.org/tasks/browse/OOIION-645)


blazetopher:
:+1:

PR #503, opened 2013-01-25
Patches visual retrieve to use sys-time

lukecampbell:
The patch earlier to improve performance with retrieve does internal
conversion from system time to dataset time units, this patches the
visual call that was exploiting a bug in retrieval.

Part of [OOIION-645](https://jira.oceanobservatories.org/tasks/browse/OOIION-645)


blazetopher:
:+1:

PR #507, opened 2013-01-28
Adds int test to verify retrieve times are correct.

lukecampbell:
Part of [OOIION-695](https://jira.oceanobservatories.org/tasks/browse/OOIION-695)

### DIFF
--- a/ion/services/dm/test/test_dm_end_2_end.py
+++ b/ion/services/dm/test/test_dm_end_2_end.py
@@ -34,6 +34,7 @@ from gevent.event import Event
 from nose.plugins.attrib import attr
 
 import gevent
+import time
 import numpy as np
 import os
 import unittest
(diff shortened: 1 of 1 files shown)


ion/services/dm/test/test_dm_end_2_end.py
```diff
@@ -528,6 +529,27 @@ def test_repersist_data(self):
 
     @attr('LOCOINT')
     @unittest.skipIf(os.getenv('CEI_LAUNCH_TEST', False), 'Host requires file-system access to coverage files, CEI mode does not support.')
+    def test_correct_time(self):
+
+        unix_now = np.floor(time.time())
+        ntp_now  = unix_now + 2208988800
```
daf: Can you put a comment in saying what these magic constants are about?

daf:
Looks fine other than previous comment.

lukecampbell:
Updated, for all following the thread: there are exactly 2208988800 seconds between 1/1/1900 and 1/1/1970, which is the conversion factor used to convert unix timestamps into NTP timestamps as per RFC 5905

PR #516, opened 2013-01-30
Adds computed attribute wrapper to UNS

lukecampbell:
The UNS standard service call to get_subscriptions returns a python
list, but certain extended attributes depend on this value being in a
computed attributes container. The new method calls get_subscriptions
and wraps the return value in the appropriate container.

Part of [OOIION-654](https://jira.oceanobservatories.org/tasks/browse/OOIION-654)


blazetopher:
:+1:

PR #518, opened 2013-01-30
Improves temporal bounds for large datasets.

lukecampbell:


blazetopher:
Should help vis performance! :+1:

PR #525, opened 2013-01-31
Fixes buildbot: test RMS on a less restrictive resource type that still has a lifecycle

ianfixes:


mmeisinger:
Looks good. Thanks

PR #526, opened 2013-01-31
Parameter Definition Consolidation

lukecampbell:
Part of [OOIION-579](https://jira.oceanobservatories.org/tasks/browse/OOIION-579).

Consolidates the schema for defining additional parameters and organization for the various parameter, parameter types and collections, known as Parameter Dictionaries. 
# Adding Parameters

All parameters to be used for datasets are _currently_ defined in the Google Docs 
[R2 Preload Resources](https://docs.google.com/spreadsheet/ccc?key=0AttCeOvLP6XMdG82NHZfSEJJOGdQTkgzb05aRjkzMEE#gid=57) 
spread sheet. Under the tab for ParameterDefs, authors can define additional parameters to be used by the ION system.
## Columns
- **Scenario**: This column declares the group that the parameter is a part of. When the system loads resources from
  the spreadsheet, they are loaded based on their scenario. Scenario defines the order and the group of resources to
  load at any given time. 
- **confluence**: Any reference material pertaining to the parameter defined in the confluence documentation should be
  listed in this column; specifically, this refers to data product specifications.
- **Name**: The canonical name for the parameter. Names should practice the programming practice of being underscore
  '_' delimited. Example names include: `battery_voltage_main` `pump_mode` `temperature` `sound_velocity`
- **ID**: **This column must be unique**. The ID column contains a unique identifier used to reference the parameter in
  other sheets of the spreadsheet. Resource identification numbers are generated at runtime (when the system actually
  loads them), and they are random, so in order to refer to parameters in the spreadsheet, we use the ID columns.
- **ID Conflict**: Identifies a conflicting ID number, not to be filled in.
- **HID**: A deprecated column that shows the parameter and pertinent information in an underscore delimited format. Do
  not fill this in.
- **HID Conflict**: No longer used
- **ParameterType**: The computational type that the parameter is, described in the parameter type section.
- **Value Encoding**: The primitive characterization for the parameter, described below in the paramater type section.
- **Code Set**: Only used if the parameter type is a `category<>`, described in the parameter type section.
- **Unit of Measure**: The UDUnits system to describe the units that the parameter measures.
- **Fill Value**: A value to represent the absence of a value. This lets operators and programs to understand that the
  parameter's value is missing or invalid. Note: `NaN` is not allowed to be a fill value.
#### QuantityType

A numerical primitive representation, this is the most widely used parameter type. This type is a container for numeric
values and is optimized for various types of numeric values. To use this parameter type use `quantity` in the Parameter
Type column and fill in the appropriate value encoding and fill value.

Appropriate value encodings are:
- Integer types: `int8 int16 int32 int64`
- Unsigned Integers (positive values only): `uint8 uint16 uint32 uint64`
- Floating Point Values: `float32 float64`
#### Array

A single record that holds multiple values, the values could be either a string of characters or multiple numeric
values. To represent an array use `array<` _parameter type_ `>` where the value inside the angle brackets are the type of
the values inside each array. The two most common exampes are `array<str>` to represent a variable length string or
`array<quantity>` to represent an array of values. The value encoding must be appropriate for the inside parameter type,
like `float32` for an `array<quantity>` or `str` for `array<str>`.
#### Boolean

A record that is either true or false. Parameters of this type should be defined as `boolean`. Conditional expressions
and boolean values are actually numeric expressions where any non-zero value is considered true and zero is false. Value
encoding is not needed for these parameter types but the fill value should be either `true` or `false`.  
#### Range

A record that represent a range of numeric values. The appropriate parameter type for this is `range<quantity>`. 
#### Constants

Two specific constants are used, a numeric constant `constant<quantity>` or a constant string `constant<str>`. If a
constant string is used a **valid length must be specified in the value encoding column**: the length is defined by
using 'S' followed by the length.  Examples: S8 represents an 8 character string, S12 a 12 character string etc.
Standard practice is to leave a little extra room, so if you know the string takes up only 11 characters make it S16
just to leave room for error and growth.
#### Record

Represents a complex value composed of Python primitives, typically a dictionary.
#### Category

This is an enumerated type, internally represented as an integer but externally represented as a meaningful string. They
are defined similar to C enumeration types. The proper parameter type is `category<` _integer type_ `:str>` and the code
set defines the enumeration, the value encoding should be an integer type, `int8` is the common. An example is:

```
Parameter Type: category<int8:str>
Value Encoding: int8
Code Set: {0: 'apples', 1:'oranges', 2:'bananas', -9: 'no fruit'}
Fill Value: -9
Unit of Measure: 1
```

### DIFF
--- a/ion/processes/bootstrap/ion_loader.py
+++ b/ion/processes/bootstrap/ion_loader.py
@@ -42,13 +42,6 @@
 
 __author__ = 'Michael Meisinger, Ian Katz, Thomas Lennan'
 
-import ast
-import calendar
-import csv
-import numpy as np
-import re
-import requests
-import time
 
 from pyon.core.bootstrap import get_service_registry
 from pyon.core.exception import NotFound
@@ -61,18 +54,26 @@ from ion.core.ooiref import OOIReferenceDesignator
 from ion.processes.bootstrap.ooi_loader import OOILoader
 from ion.processes.bootstrap.ui_loader import UILoader
 from ion.services.dm.utility.granule_utils import time_series_domain
+from ion.services.dm.utility.types import get_parameter_type, get_fill_value
 from ion.agents.port.port_agent_process import PortAgentProcessType, PortAgentType
-from ion.util.parameter_loader import ParameterPlugin
 from ion.util.xlsparser import XLSParser
 from coverage_model.parameter import ParameterContext
 from coverage_model.parameter_types import QuantityType, ArrayType, RecordType
 from coverage_model.basic_types import AxisTypeEnum
-from ion.util.parameter_loader import ParameterPlugin
 from ion.agents.platform.oms.oms_client_factory import OmsClientFactory
 
 
 from interface import objects
+
 import logging
+import simplejson as json
+import ast
+import calendar
+import csv
+import numpy as np
+import re
+import requests
+import time
 
 # format for time values within the preload data
 DEFAULT_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S"
@@ -89,7 +90,7 @@ CANDIDATE_UI_ASSETS = 'https://userexperience.oceanobservatories.org/database-ex
 MASTER_DOC = "https://docs.google.com/spreadsheet/pub?key=0AttCeOvLP6XMdG82NHZfSEJJOGdQTkgzb05aRjkzMEE&output=xls"
 
 ### the URL below should point to a COPY of the master google spreadsheet that works with this version of the loader
-TESTED_DOC = "https://docs.google.com/spreadsheet/pub?key=0AgkUKqO5m-ZidEZOS29JWG5EWHJkTldzVmdmV2RUX2c&output=xls"
+TESTED_DOC = "https://docs.google.com/spreadsheet/pub?key=0AgGScp7mjYjydHBEVnM1d2tIUDUtOWZNSElxaVEySWc&output=xls"
 #
 ### while working on changes to the google doc, use this to run test_loader.py against the master spreadsheet
 #TESTED_DOC=MASTER_DOC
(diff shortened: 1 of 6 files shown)


ion/processes/bootstrap/ion_loader.py
```diff
@@ -1295,70 +1296,107 @@ def _load_StreamDefinition(self, row):
...
+
+
+
+    def _load_ParameterDefs(self, row):
+        try:
+            if row['SKIP']:
+                self._conflict_report(row['ID'], row['Name'], row['SKIP'])
+                return
+        except:
+            print '>>>>>>>>>> ', row.keys()
```
lukecampbell, line 1347: I will get rid of this.

ion/services/dm/utility/test/test_types.py
```diff
@@ -0,0 +1,433 @@
...
+        [self.assertEquals(paramval[i], {0:'a'}) for i in xrange(20)]
+
+        self.rdt_to_granule(context, [{0:'a'}] * 20)
+
+    def test_bad_ptype(self):
+        self.assertRaises(TypeError, get_parameter_type, 'flimsy','','')
+
+
+@attr('EXHAUSTIVE')
+class ExhaustiveParameterTest(IonIntegrationTestCase):
```
lukecampbell, line 297: This test loads every parameter dictionary and parameter definition, builds an ion system dataset, writes values to it and retrieves them in a full DM end 2 end test.

ion/services/dm/utility/test/test_types.py
```diff
@@ -0,0 +1,433 @@
...
+            return ptype.categories.keys()[0]
+        else:
+            return
+
+
+    def wait_until_we_have_enough_granules(self, dataset_id='',data_size=40):
+        '''
+        Loops until there is a sufficient amount of data in the dataset
+        '''
+        done = False
```
lukecampbell, line 354: could use a `gevent.event.Event` here.

PR #529, opened 2013-02-04
Reduces HDF flush frequency

lukecampbell:
Just some background for those reading the PR:
[Latency Spreadsheet](https://docs.google.com/spreadsheet/ccc?key=0AgGScp7mjYjydGRtNHJMUWNsZkdMSnhTMFFwdXptaEE#gid=0)

Before, the coverage would flush regardless if something had changed, which is fine from a pure HDF perspective, it's faster to flush in HDF than it is to check if something has changed in Python code. Unfortunately HDF operations block the Python gevent context manager (even though the GIL is freed by the h5py library). This is by far the lesser of two evils and the performance is still within tolerance.


blazetopher:
:+1:

PR #533, opened 2013-02-05
Adds assertions in test_activate_instrument.py to check that subscriptions for two different users

swarbhanu:
- We verify the computed attributes holding the subscription in extended product and extended instrument. We check that when we create notifications for another user, there is no mix up and the computed attributes are correct.
- Since UNS uses elastic search to fetch subscriptions in the methods, get_subscriptions() and get_subscriptions_attribute(), we need to turn on elastic search in the test, test_activate_instrument, to completely verify the correct computation of the computed attribute for user subscriptions.

We do this using a config. We set config.bootstrap.use_es = True, and pass this config when starting the container with r2deploy.

```
        config = DotDict()
        config.bootstrap.use_es = True
        self.container.start_rel_from_url('res/deploy/r2deploy.yml', config)
```
- We cleanup elastic search indexes using addCleanup()

### DIFF
--- a/ion/processes/data/transforms/event_alert_transform.py
+++ b/ion/processes/data/transforms/event_alert_transform.py
@@ -324,6 +324,7 @@ class AlertTransformAlgorithm(SimpleGranuleTransformFunction):
 
         log.debug("Values unravelled: %s" % values)
         log.debug("Time names unravelled: %s" % time_names)
+        log.debug("Transform got the rdt here: %s", rdt)
 
         indexes = [l for l in xrange(len(time_names))]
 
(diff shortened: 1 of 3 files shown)


jamie-cyber1:
This might not work in the pycc mode.

lukecampbell:
To my knowledge coi_pycc's buildbot doesn't use ES. Please mark it with LOCOINT and the CEI skip annotation.

swarbhanu:
yes, using the LOCOINT and CEI skip notations to resolve the matter of coi_pycc not supporting ES.

lukecampbell:
It's not that it doesn't support it, it's just not being used.

PR #534, opened 2013-02-05
Removes ingestion buffering

lukecampbell:
Part of [OOIION-706](https://jira.oceanobservatories.org/tasks/browse/OOIION-706)


blazetopher:
:+1:

PR #536, opened 2013-02-05
Add a test for process launched twice by PD.

oldpatricka:
Also test that cleanup works


labisso:
Reviewed. This is part of OOIION-705

PR #540, opened 2013-02-06
Cov handler

mbarry02:


ion/util/pydap/handlers/coverage/coverage_handler.py
```diff
@@ -7,140 +7,191 @@
...
-    #extensions = re.compile(r"^.*_cov", re.IGNORECASE)
-    extensions = re.compile(r'^.*[0-9A-Za-z]{32}', re.IGNORECASE)
+    extensions = re.compile(r'^.*[0-9A-Za-z\-]{36}',re.IGNORECASE)
 
     def __init__(self, filepath):
         self.filepath = filepath
+    
+    def is_str(self, data):
+        for d in data:
+            if not isinstance(d, str):
```
lukecampbell: can you use `basestring` here instead of `str` so that we include things like bytes, unicode etc.

ion/util/pydap/handlers/coverage/coverage_handler.py
```diff
@@ -7,140 +7,191 @@
...
+                        keys = numpy.array(keys)
+                        ttype = keys.dtype.char
+                        if self.is_str(keys) == True:
+                            ttype = 's'
+                        seq[name+"_keys"] = BaseType(name=name+"_keys", data=keys, type=ttype, attributes=attrs)
+                        values = numpy.array(values)
+                        ttype = values.dtype.char
+                        if self.is_str(values) == True:
+                            ttype = 's'
+                        seq[name+"_values"] = BaseType(name=name+"_values", data=values, type=ttype, attributes=attrs)
```
lukecampbell: Can we split the logic here for all the if isinstances into a separate function which calls separate functions so that it's more manageable?  Example: https://github.com/ooici/coi-services/blob/master/ion/services/dm/utility/types.py#L88-L106

ion/util/pydap/handlers/coverage/coverage_handler.py
```diff
@@ -7,140 +7,191 @@
...
+        
+        return slice_
+    
+    def transform_categories(self, categories, cat_data):
+        result = []
+        for d in cat_data:
+            for k,v in categories.iteritems():
+                if d == v:
+                    result.append(k)
+        return result
```
lukecampbell:
```
result = [k for k,v in categories.iteritems() if v in cat_data]
```

ion/util/pydap/handlers/coverage/coverage_handler.py
```diff
@@ -7,140 +7,191 @@
...
+
+        
+    def transform_boolean(self, boolean_values):
+        result = []
+        for b in boolean_values:
+            if b == True:
+                result.append(1)
+            else:
+                result.append(0)
+        return result
```
lukecampbell:
```
result = [1 if b else 0 for b in boolean_values]
```

lukecampbell:
```
result = numpy.asanyarray(boolean_values, dtype='int8')
```

blazetopher: And if for some reason pydap doesn't like an int8 numpy array,

result = numpy.asanyarray(boolean_values, dtype='int8').tolist()

ion/util/pydap/handlers/coverage/coverage_handler.py
```diff
@@ -7,140 +7,191 @@
...
                     try:
-                        grid[dim] = get_var(coverage,dim,slice_)
+                        attrs['units'] = pc.uom
                     except:
-                        log.exception('Problem reading coverage %s', coverage.name)
-                        continue
+                        pass
+                    attrs['long_name'] = pc.long_name
+                    data = self.cov.get_parameter_values(name, tdoa=slice_)
+                    data = numpy.array(data)
```
blazetopher: Be aware that you're making a copy of the data array by doing this, which I don't believe it's necessary.  If this is a catch to make sure everything is a numpy array (which everything coming from get_parameter_values is anyhow - unless is a single value) - use this instead:

data = np.asanyarray(data)

ion/util/pydap/handlers/coverage/coverage_handler.py
```diff
@@ -7,140 +7,191 @@
...
+            if not isinstance(d, (list,tuple)):
+                return False
+        return True
 
     def parse_constraints(self, environ):
+        import sys
         base = os.path.split(self.filepath)
-        coverage = SimplexCoverage.load(base[0], base[1],mode='r')
-        #coverage = SimplexCoverage.pickle_load(self.filepath)
+        self.cov = SimplexCoverage.load(base[0], base[1],mode='r')
```
blazetopher: Is there a particular reason you're making this a class variable?  OK if so, just curious.

ion/util/pydap/handlers/coverage/coverage_handler.py
```diff
@@ -7,140 +7,191 @@
...
+                    if isinstance(pc.param_type, QuantityType):
+                        seq[name] = BaseType(name=name, data=data, type=data.dtype.char, attributes=attrs)
+                    if isinstance(pc.param_type, ConstantType):
+                        seq[name] = BaseType(name=name, data=data, type=data.dtype.char, attributes=attrs)
+                    if isinstance(pc.param_type, ConstantRangeType):
+                        try:
+                            col1,col2 = zip(*data)
+                        except:
+                            col1 = [d for d in data]
+                            col2 = [d for d in data]
```
blazetopher: How come col1 & col2 are the same thing here?  if zip blows up, what are you expecting data to be?

mbarry02: the data is a list of None's in that case which makes it work otherwise it would be an empty list which would not correspond to the dimension of the timesteps

ion/util/pydap/handlers/coverage/coverage_handler.py
```diff
@@ -7,140 +7,191 @@
...
+                        col1 = numpy.array(col1)
+                        seq[name_col1] = BaseType(name=name_col1, data=col1, type=col1.dtype.char, attributes=attrs)
+                        col2 = numpy.array(col2)
+                        seq[name_col2] = BaseType(name=name_col2, data=col2, type=col2.dtype.char, attributes=attrs)
+                    if isinstance(pc.param_type,BooleanType):
+                        boolean_values = self.transform_boolean(data)
+                        boolean_values = numpy.array(boolean_values)
+                        seq[name] = BaseType(name=name, data=boolean_values, type=boolean_values.dtype.char, attributes=attrs)
+                    if isinstance(pc.param_type,CategoryType):
+                        cat_data = self.transform_categories(pc.param_type.categories, data)
```
blazetopher: What are you transforming?  Don't the values come back as the 'values' of the dict, not the keys?

cov.get_parameter_values('category_param')
np.array('dog', 'cat', dog', 'duck', ...)

ion/util/pydap/handlers/coverage/coverage_handler.py
```diff
@@ -7,140 +7,191 @@
...
+                        cat_data = numpy.array(cat_data)
+                        bt = BaseType(name=name, data=cat_data, type=cat_data.dtype.char, attributes=attrs)
```
blazetopher: This question applies elsewhere as well - but when making a char BaseType, don't you need to specify how many characters?  or does it automagically do that based on the data you supply?

ion/util/pydap/handlers/coverage/coverage_handler.py
mbarry02: at first i thought pydap couldn't handle strings so i changed it to the enumeration
but now i could leave it as is

ion/util/pydap/handlers/coverage/coverage_handler.py
mbarry02: it handles strings...run the integration tests....i hooked it up to a pydap server

ion/util/pydap/handlers/coverage/coverage_handler.py
```diff
@@ -7,140 +7,191 @@
...
-    
-    
-    log.debug( 'name=%s', name)
-    log.debug( 'data=%s', data)
-    log.debug( 'shape=%s' , data.shape)
-    log.debug( 'type=%s', data.dtype.char)
-    log.debug( 'dims=%s', dims)
-    log.debug( 'attrs=%s', attrs)
-    return BaseType(name=name, data=data, shape=data.shape, type=data.dtype.char, dimensions=dims, attributes=attrs)
+    def update_slice_object(self, slice_, fill_index):
```
blazetopher, line 154: You might take a look at coverage_model.utils.fix_slice()  You look to be doing similar things, but that one's a little more comprehensive....which may also make it too heavy weight...comare/contrast...

mbarry02, line 154: i copied that code from previous version....some of it does not look like it runs when i tested it

lukecampbell, line 154: This is the slicing that Tim G and I came up with to accomodate for missing values from time domain.

PR #543, opened 2013-02-07
Fix for OOIION-636 - Patches for the scenario where we have multiple uns instances 

swarbhanu:
- Just as the notification workers do, a subscriber in UNS now listens
  for the reload user info event that gets published anytime a create, update
  or delete method is used in UNS. The callback used by the subscriber reloads
  the user info dictionaries from the resource registry the same way the 
  notification workers do. This ensures that all instances of the UNS can reload the
  user info from the resource registry in a common way whenever there has been a change in the notifications.

Now the notification workers and the UNS have a common framework for reloading the user info dictionaries.


daf:
Seems fine to me.  With the nature of git, you don't need to keep anything around in commented out form - you can always check out an old version of the file and extract from there.

swarbhanu:
cool. Yes, removed the commented out portion. It makes things cleaner.

PR #553, opened 2013-02-11
Fixes OOIION-714 (Data process not ready in a light or full launch test)

swarbhanu:
Checks to see that the created data process is actually running, before continuing further down the test.

### DIFF
--- a/ion/services/sa/process/test/test_int_data_process_management_service.py
+++ b/ion/services/sa/process/test/test_int_data_process_management_service.py
@@ -26,14 +26,10 @@ from interface.services.dm.ipubsub_management_service import PubsubManagementSer
 from interface.services.dm.iuser_notification_service import UserNotificationServiceClient
 from interface.objects import LastUpdate, ComputedValueAvailability
 from ion.services.dm.utility.granule_utils import time_series_domain
-from ion.services.cei.process_dispatcher_service import ProcessStateGate
 from interface.objects import ProcessStateEnum
 from mock import patch
-from coverage_model.coverage import GridDomain, GridShape, CRS
-from coverage_model.basic_types import MutabilityEnum, AxisTypeEnum
 from ion.agents.port.port_agent_process import PortAgentProcessType, PortAgentType
 import gevent
-from gevent.event import AsyncResult
 from sets import Set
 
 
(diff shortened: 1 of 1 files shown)


jamie-cyber1:
If create_data_process is already spawning a new process.  Checking the event afterwards will miss the event for a super fast spawn.

ion/services/sa/process/test/test_int_data_process_management_service.py
```diff
@@ -370,7 +369,27 @@ def test_createDataProcessUsingSim(self):
...
+            recent_events = uns.get_recent_events(resource_id=process_id)
+            for evt in recent_events.value:
+                cond = evt.type_ == 'ProcessLifecycleEvent' and evt.origin_type == 'DispatchedProcess'
+
+                # if it is the right event, check the state to see if it is running
+                if cond:
+                    log.debug("came here for the event: %s", evt.state)
+                    return evt.state == ProcessStateEnum.RUNNING
+
+        poll(data_process_running, data_process.process_id)
```
swarbhanu, line 392: Jamie, right after the create_data_process(), I have polling and use the UNS service's method get_recent_events() to check for recent events to see if an event has recently arrived that would say that the process is running. After a timeout of 10 seconds, if the poller finds that no event has appeared in the system that states that the process is not running, then there is a timeout error. On the other hand, whether for fast or slow process spawning, as soon as the poller discovers that the process is running, the poller will exit and the test will run as usual. 

So I think, there will not be any missing of event. This is because, we are not using an event subscriber. We are using UNS to do a get_recent_events() right after the create_data_process() has been executed.

swarbhanu:
Right after the create_data_process(), I do polling and use the UNS service's method get_recent_events() to check for recent events to see if an event has recently arrived that would say that the process is running. After a timeout of 10 seconds, if the poller finds that no event has appeared in the system that states that the process is not running, then there is a timeout error. On the other hand, whether for fast or slow process spawning, as soon as the poller discovers that the process is running, the poller will exit and the test will run as usual.

So I think, there will not be any missing of event. This is because, we are not using an event subscriber. We are using UNS to do a get_recent_events() right after the create_data_process() has been executed.

jamie-cyber1:
Sounds good.

PR #559, opened 2013-02-12
Added autosample_duration to InstrumentDeviceComputedAttributes for use in InstrumentDeviceExtension

swarbhanu:
- Bumps ion defs
- Implements get_autosample_duration() and get_uptime(). These values are filled after querying the events db.
- Basic integration test code added to check that the autosample_duration and uptime attributes are indeed getting computed and added to the extended resource for instrument device (InstrumentDeviceExtension) and that they are of the correct type.


swarbhanu:
A new pull request to be made soon.

PR #560, opened 2013-02-12
Refactors find_events() of UNS so that it simply calls find_events() in pyon

swarbhanu:


shenrie:
I think I closed this PR by mistake..sorry about that. However, I am not sure this operation is needed anymore if all it does is call the find_events() function in the container. We should confirm that this still is necessary....

swarbhanu:
Hi Stephen, Thats okay. Well, the pull request was asked by Michael and Maurice. Michael wanted the method in the UNS layer to call the one in the pyon layer. I will reopen another one so that it can be discussed more. Thanks,
Swarbhanu.

PR #564, opened 2013-02-13
UNS' find_events() now uses the container's event_repository's find_events() 

swarbhanu:
- Did the refactor: UNS' find_events() now uses the container's event_repository's find_events() 
- Made a test that was sporadically failing more robust by increasing the number of polling tries.


mmeisinger:
Looks good to me satisfying the goal. Thanks a lot.

PR #567, opened 2013-02-14
Should fix the r2_light and r2_full tests in buildbot

swarbhanu:
Uses an AsyncResult() to block and wait until the process is running before the rest of the main test thread is executed.

### DIFF
--- a/ion/services/sa/process/test/test_int_data_process_management_service.py
+++ b/ion/services/sa/process/test/test_int_data_process_management_service.py
@@ -30,6 +30,7 @@ from interface.objects import ProcessStateEnum
 from mock import patch
 from ion.agents.port.port_agent_process import PortAgentProcessType, PortAgentType
 import gevent
+from gevent.event import AsyncResult
 from sets import Set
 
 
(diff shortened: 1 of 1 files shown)


lukecampbell:
The `AsyncResult` is not necessary here. The `poll` as defined [here](https://github.com/ooici/pyon/blob/master/pyon/util/poller.py) will automatically block until the callback returns True or the timeout is exceeded in which case a `Timeout` is raised and the test will fail. The AsyncResult is doing the exact same utility as Poll already did.

ion/services/sa/process/test/test_int_data_process_management_service.py
```diff
@@ -379,17 +380,22 @@ def test_createDataProcessUsingSim(self):
...
             for evt in recent_events:
                 log.debug("Got an event with event_state: %s. While ProcessStateEnum.RUNNING would be: %s", evt.state, ProcessStateEnum.RUNNING)
                 if evt.state == ProcessStateEnum.RUNNING:
+                    ar.set(True)
                     return True
             return False
 
         poll(data_process_running, self.event_repo, data_process.process_id)
 
+        ar.wait(10)
```
lukecampbell, line 397: If a `Timeout` wasn't raised up to this point `data_process_running` returned True.

swarbhanu:
yes, true. but the r2_light test still fails here instead of raising a timeout as expected because poll should block. In the failing build, timeout is not getting raised instead the assertion of process running is failing. maybe something simple is getting overlooked. Otherwise, it seems to me that it is not actually running synschronously. Would be happy to know where the fail is coming from.

lukecampbell:
If you think that poll is somehow either not executing at all or not running synchronously try timing the execution time of it. There is no gevent spawning happening inside `poll`.  Poll can't return without a timeout or the callback returning true, if gevent were say broken and couldn't be used on the build then it would just loop forever, if the test is failing I can't conceive of how this AsyncResult could possibly do anything.

lukecampbell:
After reviewing the test, the assertion made [here](https://github.com/ooici/coi-services/blob/master/ion//services/sa/process/test/test_int_data_process_management_service.py#L414) is after several things have happened, it's quite possible on a distributed environment that the process state was altered between  lines 413 and line 391.

swarbhanu:
well, poll works well and this particular test works well in the master code in all the other buildbot tests (coi-nightly,coi_testbc). It only fails in r2_light. So if r2_light (at 2 am today morning) has all the latest code in master, then I am not sure why there is a failing assertion instead of the timeout.  Unless as what I think you mentioned above, maybe something happening in the r2_light and r2_full environments.

swarbhanu:
it also fails in r2_full, done also about 2 am today morning. Apart from these two test environments, all the other buildbot tests are okay with this test method.

lukecampbell:
Put a try/except around the assertion and if an assertion is raised have it dump the events surrounding the process, I bet you that it went from RUNNING to whatever PENDING and the events should reflect that.

lukecampbell:
The other builds aren't doing a distribution like these two buildbots.

swarbhanu:
Actually, the test as in master right now works in my local machine too in both pycc and without it. So, I am not able to recreate the fail in my local machine either.

PR #568, opened 2013-02-14
Granule time

tgiguere:
Adds creation_timestamp and assert to to_granule and load_from_granule in RecordDictionaryTool.

### DIFF
--- a/ion/services/dm/utility/granule/record_dictionary.py
+++ b/ion/services/dm/utility/granule/record_dictionary.py
@@ -49,6 +49,7 @@ class RecordDictionaryTool(object):
     _shp         = None
     _locator     = None
     _stream_def  = None
+    _creation_timestamp = None
     _dirty_shape = False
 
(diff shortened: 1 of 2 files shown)


lukecampbell:
:+1:

ion/services/dm/utility/granule/record_dictionary.py
```diff
@@ -118,6 +123,7 @@ def to_granule(self, data_producer_id='',provider_metadata_update={}):
         granule.domain = self.domain.shape
         granule.data_producer_id=data_producer_id
         granule.provider_metadata_update=provider_metadata_update
+        granule.creation_timestamp = time.time()
```
mmeisinger, line 126: We should use some ION time library here. We should either use the science timestamp (ntp4 derived I believe) or the ION system timestamping mechanism. I don't think we should proliferate direct time setting using time.time() or similar anywhere

PR #576, opened 2013-02-16
Ensure that EE Agent is listening before sending heartbeats

oldpatricka:
This prevents a race where the eeagent sends its first hearbeat before the
container has set up its listeners, and then the PD quickly responds, but the
message is lost, because there is no queue set up yet.


oldpatricka:
For @labisso to review when he is back from holiday.

PR #577, opened 2013-02-18
Delay coverage creation

lukecampbell:
- DataProduct's only persist a Coverage object to the file system in the DEPLOY/AVAILABLE state. 
- Computed attributes are updated to reflect new queries.
- DataProductManagement's `create_data_product` no longer accepts a raw parameter dictionary as a parameter.


blazetopher:
:+1:

PR #578, opened 2013-02-19
Support creating resource attachments in service gateway

daf:
This custom route (POST to `/ion-service/attachment`) mirrors the GET side of getting an attachment.

Also adds a change to logging of incoming requests for more debuggability.

### DIFF
--- a/ion/services/coi/service_gateway_service.py
+++ b/ion/services/coi/service_gateway_service.py
@@ -629,7 +629,7 @@ def get_attachment(attachment_id):
 def create_attachment():
 
     try:
-        resource_id        = str(request.form['resource_id'])       # unicode coming in
+        resource_id        = str(request.form.get('resource_id', ''))       # unicode coming in
         fil                = request.files['file']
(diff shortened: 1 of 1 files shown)


daf:
Good points.  What I added here supports the functionality from the UI which is create and read only.  I can add a piece to handle a resource_id not being set.  IMO an update would be a `PUT` which should be a separate handling method, and I don't think we need to add one yet until something needs it.

shenrie:
One additional thought... I saw your comment about the resource_id coming in as unicode; you can see in other areas of the service gateway that I call convert_unicode(resource_id) for example and you might want to do the same thing since unicode strings are not valid for service operation calls.... or like this... convert_unicode(request.args['requester'])

daf:
Looks like most of your comment was cut off, but I take it you are referring to the `convert_unicode` method [here](https://github.com/ooici/coi-services/blob/master/ion/services/coi/service_gateway_service.py#L521-L529).  I'll see if it works for this flow.

PR #582, opened 2013-02-21
Adds input validation and fill_value to rdt

lukecampbell:
Input vectors into the record dictionary tool will replace None with an
appropriate fill value for the specified parameter.
RecordDictionaryTool now as a fill_value method to determine the fill
value for a specified parameter.


blazetopher:
:+1:

PR #604, opened 2013-02-26
Several changes and additions to platform agent framework

carueda:
Various refactorings (mostly internal); code clean-ups; MONITORING state introduced; consistent use of a supporting NetworkDefinition object currently used to configure the platform agent while appropriate mechanism are defined; added commands CONNECT_INSTRUMENT, DISCONNECT_INSTRUMENT and GET_CONNECTED_INSTRUMENTS; renaming in CI-OMS interface to follow general CI coding guidelines; platform driver FSM introduced; added utilities to load networkDefinition from info provided by CI-OMS interface (used in various tests, including test_oms_launch2); preliminary checksum implementation and related platform agent command CHECK_SYNC; preliminary preparations for the CGSN case. Also, for better naming consistency, renamed the RSN platform module and class as follows:
   ion.agents.platform.oms -> ion.agents.platform.rsn
   ion.agents.platform.rsn.oms_platform_driver -> ion.agents.platform.rsn.rsn_platform_driver
   OmsPlatformDriver -> RsnPlatformDriver
So, per https://confluence.oceanobservatories.org/display/CIDev/Preload+development, I made a copy of the R2 Preload Resources spreadsheet, adjusted the PlatformAgent sheet to reflect the new pa/driver module and pa/driver class columns, and adjusted ion_loader.py accordingly.  

**NOTE**: Because of the above renaming in the RSN driver module and class, we will probably need to coordinate at the time of merging so a new copy of the original spreadsheet is used to reflect its most recent updates. Please let me know how you want to proceed.


ion/agents/platform/cgsn/cgsn_client.py
```diff
@@ -0,0 +1,126 @@
...
+import logging
+
+from gevent.socket import *
+from gevent import Greenlet, select
+
+from ion.agents.platform.cgsn.defs import MessageIds, CIPOP, CICGINT, EOL
+
+
+class CgsnClient(object):
+    """
```
MauriceManning, line 25: naming convention, should be CGSNClient, same for factory class below

ion/agents/platform/cgsn/cgsn_client.py
```diff
@@ -0,0 +1,126 @@
...
+        self._recv.set_listener(listener)
+
+    def start(self):
+        self._recv.start()
+
+    def end_reception(self):
+        self._recv.end()
+
+    def _sendto(self, data):
+        if log.isEnabledFor(logging.DEBUG):
```
MauriceManning, line 56: why is this necessary? doesnt the logging framework check this before logging any line that is log.debug?

ion/agents/platform/cgsn/cgsn_client_factory.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+from pyon.public import log
+import logging
+
+from ion.agents.platform.cgsn.cgsn_client import CgsnClient
+import os
+import yaml
+
+_URI_ALIASES_FILENAME = 'ion/agents/platform/cgsn/cgsn_uri_aliases.yml'
```
MauriceManning, line 21: align with how other param files are opened in the config folder

ion/agents/platform/cgsn/cgsn_uri_aliases.yml
```diff
@@ -0,0 +1,14 @@
+#
```
MauriceManning, line 1: this file should go into the res or config folder, coord with Jami. perhaps one file for CGSN configurations so cgsn.yml?

ion/agents/platform/cgsn/notes.txt
```diff
@@ -0,0 +1,21 @@
+CGSN platform agent/driver dev notes
```
MauriceManning, line 1: These and other notes.txt in this PR would be more useful in a CIDev confluence page to raise visibility and aggregated into one area versus multiple folders

ion/agents/platform/platform_agent.py
```diff
@@ -64,22 +67,49 @@
...
+    UNINITIALIZED     = ResourceAgentState.UNINITIALIZED
+    INACTIVE          = ResourceAgentState.INACTIVE
+    IDLE              = ResourceAgentState.IDLE
+    STOPPED           = ResourceAgentState.STOPPED
+    COMMAND           = ResourceAgentState.COMMAND
+    MONITORING        = 'PLATFORM_AGENT_STATE_MONITORING'
+
+
+class PlatformAgentEvent(BaseEnum):
+    ENTER                     = ResourceAgentEvent.ENTER
```
MauriceManning, line 83: I believe these should use AgentCommandEvents, see https://docs.google.com/drawings/d/1LXa33E_aBFeZxloegRQ7uMZ78lxhAmOGmL-wyMv83LM/edit and check with EH.

ion/agents/platform/cgsn/cgsn_client.py
daf, line 56: it does do the check. in the line below, he is doing string formatting before the call to debug, so presumably he was trying to save the computation time of doing that.

you can do this by using comma instead of %, as the log methods accepts *args and do formatting later for you: `log.debug("calling sendto(%r)", data)`

although there is still a few more calls that happen before it checks log level, so depending how often this i called, this check may be valid.

ion/agents/platform/platform_agent.py
MauriceManning, line 83: origin_type: type of agent resource :platform
sub_type: event specific : RESET  for example

ion/agents/platform/cgsn/cgsn_client.py
carueda, line 56: Hi guys, thanks for reviewing!
"Guarded logging" is a pretty standard practice, see for example: http://www.slf4j.org/faq.html#logging_performance, which also mentions that the overhead incurred is insignificant while at the same time making sure that potentially costly operations needed to construct the logged message are avoided when unnecessary.  Also, I'm using the % operator because the message _will_ be generated per the guarded condition.

Note, to be more clear, this standard practice actually applies in general to both statically typed and dynamic languages but that do not have a "parameters by name" or similar mechanism, like in Scala, for example.

ion/agents/platform/cgsn/cgsn_client.py
carueda, line 25: Is there a particular guideline about acronyms in CI? I don't see it in https://confluence.oceanobservatories.org/display/CIDev/Code+Conventions at least, but I'm happy to change this if required. On the other hand, AFAICT there is no definite consensus about how to handle acronyms in naming types (_) but I'm certainly in favor of using just the upper case for the first letter of the acronym because "lowercased (e.g. XmlHttpRequest) ... ease typing and readability." (_) Again happy to change this if required.

(*) see for example http://en.wikipedia.org/wiki/Naming_convention_(programming)  (search for acronym).

carueda, line 25: Thanks Steve F. for your note:

> FYI, from PEP8:
> Note: When using abbreviations in CapWords, capitalize all the letters of the abbreviation. Thus HTTPServerError is better than > > > HttpServerError.
> 
> Not sure how tight we are still sticking to PEP8, but that was the original plan.

Fair enough. I was not aware that PEP8 had this (and the checking tool does not support it --of course-; ; )

ion/agents/platform/cgsn/cgsn_client_factory.py
carueda, line 21: The location would be res/config/?  If so, I see that is a soft link:
    res -> extern/ion-definitions/res/
Would it mean the cgsn_uri_aliases.yml should be in the ion-definitions repo? Please explain, I'm bit confused.

PR #606, opened 2013-02-26
Better error handling for retrieving against corrupt coverages.

lukecampbell:
This adds some better error handling, and more verbose description of
what happens.  If a coverage's parameter's extents are larger than the
time domain, a corruption has ocurred and this extre information will
help uncover the causes.


blazetopher:
:+1:

PR #610, opened 2013-02-27
Adds deadman timeout retry algorithm to ingestion

lukecampbell:


blazetopher:
:+1:

mmeisinger:
Good pattern to increase robustness. In a distributed system, just wait and retry. The best way to eliminate "heisenbugs" is to retry

PR #634, opened 2013-03-05
Record dictionary computing function values

lukecampbell:
Record Dictionary Tool contains the capability to calculate dependent
function values.


blazetopher:
:+1:

PR #641, opened 2013-03-07
Fixes Ion-Alpha

swarbhanu:
Makes UNS handle startup scenarios gracefully so that it does not crash as a service

### DIFF
--- a/ion/services/dm/presentation/user_notification_service.py
+++ b/ion/services/dm/presentation/user_notification_service.py
@@ -466,7 +466,10 @@ class UserNotificationService(BaseUserNotificationService):
         @throws NotFound    object with specified parameters does not exist
         """
 
-        event_tuples = self.container.event_repository.find_events(event_type=type, origin=origin, start_ts=min_datetime, end_ts=max_datetime, limit=limit, descending=descending)
+        try:
+            event_tuples = self.container.event_repository.find_events(event_type=type, origin=origin, start_ts=min_datetime, end_ts=max_datetime, limit=limit, descending=descending)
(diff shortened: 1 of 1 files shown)


ion/services/dm/presentation/user_notification_service.py
```diff
@@ -466,7 +466,10 @@ def find_events(self, origin='', type='', min_datetime=0, max_datetime=0, limit=
         @throws NotFound    object with specified parameters does not exist
         """
 
-        event_tuples = self.container.event_repository.find_events(event_type=type, origin=origin, start_ts=min_datetime, end_ts=max_datetime, limit=limit, descending=descending)
+        try:
+            event_tuples = self.container.event_repository.find_events(event_type=type, origin=origin, start_ts=min_datetime, end_ts=max_datetime, limit=limit, descending=descending)
+        except Exception as exc:
+            log.warning("The UNS find_events operation for event origin = %s and type = %s failed. Error message = %s", origin, type, exc.message)
 
         events = [item[2] for item in event_tuples]
```
mmeisinger, line 474: Careful: this will NOT work. You are using the event_tuples variable in case of success and error. It will bomb if the exception is caught. I've seen this in other places too. Make sure event_tuples is initialized to an empty list before

A unit test doing branch coverage would catch this.

swarbhanu, line 474: Thanks. I made the change, initializing events_tuple as [] at the beginning of the method, and made a pull request with that:
https://github.com/ooici/coi-services/pull/646

PR #650, opened 2013-03-07
Adds Ctdbp transforms for pathfinder

swarbhanu:
- Adds the CTDBP L0, L1, L2 salinity and L2 density transforms.
- Note that there is one L0 transform, one L1 transform and two L2 transforms.
- Adds integration tests: one each for L0 and L1, and another integration test which links L0, L1 and the two L2 (density and salinity) transforms together in a chain and checks that the granules published by each is the right kind.
- For the test, we are using fictitious values of calibration coefficients. If we later know about the calibration coefficients used in the real world, maybe those could be imported later for the test. Note that these CTDBP transforms are not going to stay for long. They are only for the pathfinder. Pending changes to the transform structure, they might be replaced.
- What may be added later is more assertions to check that the algorithms have been performed. So far we are only checking that the granules published by each transform contains data for the appropriate parameters.
- Also, in test_ctd_transforms.py, there were lines that were unnecessary, so they have been removed.

### DIFF
--- a/ion/processes/data/transforms/ctdbp/ctdbp_L1.py
+++ b/ion/processes/data/transforms/ctdbp/ctdbp_L1.py
@@ -88,16 +88,16 @@ class CTDBP_L1_TransformAlgorithm(SimpleGranuleTransformFunction):
 
         # Set the conductivity values for the output granule
         # Note that since the conductivity caculation depends on whether TEMPWAT_L1, PRESWAT_L1 have been calculated, we need to do this last
-        out_rdt = CTDBP_L1_TransformAlgorithm.calculate_conductivity(   out_rdt = out_rdt,
-                                                                        cond_calibration_coeffs = cond_calibration_coeffs,
-                                                                        TEMPWAT_L1 = TEMPWAT_L1,
-                                                                        PRESWAT_L1 = PRESWAT_L1)
+        out_rdt = CTDBP_L1_TransformAlgorithm.calculate_conductivity(   input_rdt = rdt,
+                                                                        out_rdt = out_rdt,
+                                                                        cond_calibration_coeffs = cond_calibration_coeffs
+        )
 
         # build the granule for the L1 stream
         return out_rdt.to_granule()
 
     @staticmethod
-    def calculate_conductivity(input_rdt = None, out_rdt = None, cond_calibration_coeffs = None, TEMPWAT_L1=None, PRESWAT_L1=None):
+    def calculate_conductivity(input_rdt = None, out_rdt = None, cond_calibration_coeffs = None):
         """
         Dependencies: conductivity calibration coefficients, TEMPWAT_L1, PRESWAT_L1
         """
@@ -105,6 +105,9 @@ class CTDBP_L1_TransformAlgorithm(SimpleGranuleTransformFunction):
         log.debug("L1 transform applying conductivity algorithm")
 
         CONDWAT_L0 = input_rdt['conductivity']
+        TEMPWAT_L1 = out_rdt['temperature']
+        PRESWAT_L1 = out_rdt['pressure']
+
 
         #------------  CALIBRATION COEFFICIENTS FOR CONDUCTIVITY  --------------
         g = cond_calibration_coeffs['g']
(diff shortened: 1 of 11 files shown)


ion/processes/data/transforms/ctdbp/ctdbp_L2_salinity.py
```diff
@@ -0,0 +1,94 @@
...
+        if packet == {}:
+            return
+
+        log.debug("CTDBP L2 salinity transform received granule with record dict: %s", packet.record_dictionary)
+
+        granule = CTDBP_SalinityTransformAlgorithm.execute(packet, params=self.stream_definition._id)
+
+        log.debug("CTDBP L2 salinity transform publishing granule with record dict: %s", granule.record_dictionary)
+
+        granule.data_producer_id=self.id
```
MauriceManning: is there a reason that this is only set in the L2 transform?

ion/processes/data/transforms/ctdbp/ctdbp_L2_salinity.py
```diff
@@ -0,0 +1,94 @@
...
+
+        out_rdt['time'] = rdt['time']
+
+        conductivity = rdt['conductivity']
+        pressure = rdt['pressure']
+        temperature = rdt['temperature']
+
+        sal_value = SP_from_cndr(conductivity * 10, t=temperature, p=pressure)
+
+#        sal_value = SP_from_cndr(r=conductivity/cte.C3515, t=temperature, p=pressure)
```
MauriceManning: add a comment if this version of the calc needs to be saved so the next person working on the code knows why.

ion/processes/data/transforms/ctdbp/ctdbp_L2_salinity.py
lukecampbell: I think that originally we were going to use the data_producer_id to assist in tracking provenance but we shifted to using associations and data acquisition to track provenance. Also, when the ingestion framework was in its infancy the data_producer_id was used to determine which coverage the data belonged to before streams were readily identifiable. 

I don't believe we need to set data_producer_id anymore.

ion/processes/data/transforms/ctdbp/ctdbp_L1.py
```diff
@@ -0,0 +1,230 @@
...
+        CTcor = cond_calibration_coeffs['CTcor']
+        CPcor = cond_calibration_coeffs['CPcor']
+
+        if not (g and h and I and j and CTcor and CPcor):
+            raise BadRequest("All the conductivity calibration coefficients (g,h,I,j,CTcor, CPcor) were not passed through"
+                             "the config. Example: config.process.calibration_coeffs.cond_calibration_coeffs['g']")
+
+        #------------  Computation -------------------------------------
+        freq = (CONDWAT_L0 / 256) / 1000
+        CONDWAT_L1 = (g + h * freq**2 + I * freq**3 + j * freq**4) / (1 + CTcor * TEMPWAT_L1 + CPcor * PRESWAT_L1)
```
lukecampbell, line 132: Can you place in-comment citations for the numerical models?  Example: https://github.com/lukecampbell/pygsw/blob/master/pygsw/practical_salinity.pyx#L191

In this case it only needs to be reference either the DPS or the document that explains why we're computing it this way.

ion/processes/data/transforms/ctdbp/test/test_ctdbp_L0.py
```diff
@@ -0,0 +1,250 @@
...
+
+        sal_ctxt = ParameterContext('salinity', param_type=QuantityType(value_encoding=numpy.dtype('int32')))
+        sal_ctxt.uom = ''
+        pdict.add_context(sal_ctxt)
+
+        #create temp streamdef so the data product can create the stream
+        pc_list = []
+        for pc_k, pc in pdict.iteritems():
+            pc_list.append(self.dataset_management.create_parameter_context(pc_k, pc[1].dump()))
+
```
lukecampbell, line 121: In tests like this please use addCleanup to mop up the test resources:

``` python
ctxt_id = self.dataset_management.create_parameter_context(pc_k, pc[1].dump())
seld.addCleanup(self.dataset_management.delete_parameter_context,ctxt_id)
```

ion/processes/data/transforms/ctdbp/ctdbp_L1.py
swarbhanu, line 132: Good point. I have now added the reference of the spreadsheet with the calculations (that Chris Winguard had placed there for the ctdbp transform algorithms to use). Tim A wanted the ctdbp transforms to be based on that spreadsheet.

swarbhanu:
Added cleanups (Luke suggested) and removed a commented legacy computation algorithm. The new commits pushed in are according to  Maurice and Luke's comments.

PR #665, opened 2013-03-08
Handle mi exceptions

newbrough:
MI exceptions are now passed as a triple (code, message, stack info) instead of exception objects so that there are no dependencies between pycc and MI to have certain classes.


mmeisinger:
code looks good. Very beneficial to define an exception "communication protocol" and to instantiate exceptions on either side of the wire dynamically. Also good that both sides use the same utilities classes

PR #673, opened 2013-03-11
Ion functions

lukecampbell:
## Adds a new submodule for marine science functions

This will require a buildout to accomodate the changes. Very important component is that developers will need to include the libgswteos-10 library:

``` bash
brew tap lukecampbell/hombrew-libgswteos
brew install libgswteos-10
brew test -v libgswteos-10
```

A subsequent buildout will include all of the necessary libraries. 

This submodule will host a repository for authors to define and manage functions to execute marine science calculations. These are required to calculate the marine seawater variables used ultimately for deriving data products from instrument variables.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,3 +17,6 @@
 [submodule "extern/eeagent"]
 	path = extern/eeagent
 	url = https://github.com/ooici/eeagent.git
+[submodule "extern/ion-functions"]
+	path = extern/ion-functions
+	url = https://github.com/ooici/ion-functions.git
--- a/README
+++ b/README
@@ -47,6 +47,12 @@ For Mac, use homebrew
     (If you have trouble with brew 'MD5 mismatch' errors, try running 'brew update'
     and try the install again)
 
+- Install libgswteos-10 for marine science calculations
+    > brew tap lukecampbell/homebrew-libgswteos
+    > brew install libgswteos-10
+    > brew test -v libgswteos-10
+    
+
 
 installation
 - Install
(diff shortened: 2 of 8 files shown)


blazetopher:
Lookin' good.  Needs review and go-ahead from Michael/Maurice.
:+1:

mmeisinger:
Looks good to me. This will be a good place to author functions.

Doing it as a submodule is common practice now (lump it all into coi-services and keep coi-services commit as the master commit) and I'm fine with it. In the longer run a more detached repository would be preferable, but that's the past-R2 future. (and that's true to other submodules as well).

In order to limit the ripples for the rest of developers, we need to understand the impact. Is everyone required at the time of merge to make the brew calls? Then we need to send out an email to all. How often will everyone have to do a buildout once ion-functions are populated. It would be preferable if failure were a bit more graceful for a while, e.g. test skips if certain dependencies are not there. Merge when ready and impacts are understood.

lukecampbell:
Developers will need to install the packages (via brew if on mac) in order for their container to take advantage of these functions otherwise the computed values will simply be empty inside an RDT or an exception will be raised if attempting to get it from coverage directly. I'm writing tests now that test the flow of the preload parameters and precision of the marine functions so those tests will fail if the `libgswteos-10` package isn't installed.

Buildout should still be sparse, and we would only require one if and only if a function author needs another third-party library. I will send an email to everyone and coordinate with Jamie to make sure the buildbots are configured, once this is done I will go ahead and merge this PR.

lukecampbell:
Verified `libgswteos-10` installed on `r2-dev1`, configured and `pygsw` compiles. We are all ready!

PR #674, opened 2013-03-11
Bump SBE37 Driver Version

wfrench:
Updated the agent tests to use a new version of the sbe37 driver
removed all references to mi driver references and replaced them with their ION errors


edwardhunter:
Looks good.

PR #678, opened 2013-03-11
expanding parser for 'simple' dicts

ianfixes:
Ready to merge.  This PR adds support for more complicated 'simple' dicts.  

``` python
        But is it really simple?  Not quite.  The following substitutions are made:

        keys with dots ('.') will be split into dictionaries.
        booleans "True", "False" will be parsed
        numbers will be parsed as floats unless they begin with "0" or include one "." and end with "0"
        "{}" will be converted to {}
        "[]" will be converted to [] (that's "[ ]" with no space)

        For example, an entry in preload would be this:

        PARAMETERS.TXWAVESTATS: False,
        PARAMETERS.TXREALTIME: True,
        PARAMETERS.TXWAVEBURST: false,
        SCHEDULER.ACQUIRE_STATUS: {},
        SCHEDULER.CLOCK_SYNC: 48.2
        SCHEDULER.VERSION: 3.0

        which would translate back to
        { "PARAMETERS": { "TXWAVESTATS": False, "TXREALTIME": True, "TXWAVEBURST": "false" },
          "SCHEDULER": { "ACQUIRE_STATUS": { }, "CLOCK_SYNC", 48.2, "VERSION": "3.0"}
        }

```

In addition, startup_config is added to the preload sheet for InstrumentAgentInstance resources, imported as a "simple dict".

### DIFF
--- a/ion/processes/bootstrap/ion_loader.py
+++ b/ion/processes/bootstrap/ion_loader.py
@@ -871,7 +871,7 @@ class IONLoader(ImmediateProcess):
 
         which would translate back to
         { "PARAMETERS": { "TXWAVESTATS": False, "TXREALTIME": True, "TXWAVEBURST": "false" },
-          "SCHEDULER": { "ACQUIRE_STATUS": { }, "CLOCK_SYNC", 48.3, "VERSION": "3.0"}
+          "SCHEDULER": { "ACQUIRE_STATUS": { }, "CLOCK_SYNC", 48.2, "VERSION": "3.0"}
         }
 
         """
--- a/ion/processes/bootstrap/ion_loader.py
+++ b/ion/processes/bootstrap/ion_loader.py
@@ -849,16 +849,80 @@ class IONLoader(ImmediateProcess):
         self.contact_defs[id] = contact
 
     def _parse_dict(self, text):
-        """ parse a SIMPLE dictionary of unquoted string keys and values
         """
+        parse a "simple" dictionary of unquoted string keys and values. -- no nested values, no complex characters
+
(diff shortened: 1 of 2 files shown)


newbrough:
1) since python doesn't distinguish between int and float types, could just
parse as float and drop L885-890
2) i see no code handling the number case beginning with 0 or with "." and
ending with 0.
3) example in comment: 48.2 becomes 48.3 ?

ianfixes:
I had to add the int parsing, because str(float("3")) == "3.0"

Instead of just checking the leading zero or trailing 0 after ".", I just did the more broad test of converting only in cases where the string value of the text was not altered.

Shown here:
https://github.com/ooici/coi-services/pull/678/files#L0R887
https://github.com/ooici/coi-services/pull/678/files#L0R894

These catch the cases we talked about, and others.

ianfixes:
Ready to merge

ianfixes:
Note that this code will not work for the following example, shown here:
https://confluence.oceanobservatories.org/display/syseng/CIAD+MI+SV+Instrument+Driver+Scheduled+Tasks

``` python
startup_config = {
    DriverConfigKey.SCHEDULER: {
        job_name: {
            DriverSchedulerConfigKey.TRIGGER: {
                DriverSchedulerConfigKey.TRIGGER_TYPE: TriggerType.ABSOLUTE,
                DriverSchedulerConfigKey.DATE: dt
            }
        }
    }
}
```

The best it can do is turn `SCHEDULER.job_name: {}` into `python {'SCHEDULER': {'job_name': {}}}`

PR #688, opened 2013-03-12
Lat and lon is now sent as config to the density transform which uses the values

swarbhanu:
- Additionally real values for the calibration coefficients were used.
- The names for the calibration coefficients are aligned with the ones sent in email thread from the real instrument.

### DIFF
--- a/ion/processes/data/transforms/ctdbp/ctdbp_L2_density.py
+++ b/ion/processes/data/transforms/ctdbp/ctdbp_L2_density.py
@@ -14,6 +14,8 @@ from interface.services.dm.ipubsub_management_service import PubsubManagementSer
 from seawater.gibbs import SP_from_cndr, rho, SA_from_SP, conservative_t
 from seawater.gibbs import cte
 
+import numpy as np
+
 class CTDBP_DensityTransform(TransformDataProcess):
     """ A basic transform that receives input through a subscription,
     parses the input from a CTD, extracts the pressure value and scales it according to
@@ -30,9 +32,14 @@ class CTDBP_DensityTransform(TransformDataProcess):
                              "using a special keyword (ex: density)")
         self.dens_stream_id = self.CFG.process.publish_streams.density
 
+        lat = self.CFG.lat
+        lon = self.CFG.lon
+
         # Read the parameter dict from the stream def of the stream
         pubsub = PubsubManagementServiceProcessClient(process=self)
         self.stream_definition = pubsub.read_stream_definition(stream_id=self.dens_stream_id)
+        
+        self.params = {'stream_def' : self.stream_definition._id, 'lat': lat, 'lon' : lon}
 
     def recv_packet(self, packet, stream_route, stream_id):
         """
@@ -42,7 +49,7 @@ class CTDBP_DensityTransform(TransformDataProcess):
             return
         log.debug("CTDBP L2 density transform received granule with record dict: %s", packet.record_dictionary)
 
-        granule = CTDBP_DensityTransformAlgorithm.execute(packet, params=self.stream_definition._id)
+        granule = CTDBP_DensityTransformAlgorithm.execute(packet, params=self.params)
 
         log.debug("CTDBP L2 density transform publishing granule with record dict: %s", granule.record_dictionary)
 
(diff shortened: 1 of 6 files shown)


ion/processes/data/transforms/ctdbp/ctdbp_L2_density.py
```diff
@@ -30,9 +32,14 @@ def on_start(self):
                              "using a special keyword (ex: density)")
         self.dens_stream_id = self.CFG.process.publish_streams.density
 
+        lat = self.CFG.lat
+        lon = self.CFG.lon
```
lukecampbell, line 36: since lat and lon are specific to the proces please set them under `CFG.process.lat` and `CFG.process.lon` also when reading variables from a DotDict you should use `get_safe`

``` python
lat = self.CFG.get_safe('process.lat', 0)
lon = self.CFG.get_safe('process.lon', 0)
```

ion/processes/data/transforms/ctdbp/ctdbp_L2_density.py
```diff
@@ -83,8 +93,10 @@ def execute(input=None, context=None, config=None, params=None, state=None):
         pressure = rdt['pressure']
         temperature = rdt['temp']
 
-        longitude = rdt['lon'] if rdt['lon'] is not None else 0
-        latitude = rdt['lat'] if rdt['lat'] is not None else 0
+        latitude = np.ones(len(conductivity)) * lat
```
lukecampbell, line 96: I recommend using shape here:

``` python
latitude = np.ones(conductivity.shape) * lat
```

If we start using multi-dimensional data this will be one last component we'll have to change.
I still don't know if this is the best way to make an array with all the same values though.

ion/processes/data/transforms/ctdbp/test/test_ctdbp_L1.py
```diff
@@ -128,31 +118,52 @@ def _get_new_ctd_L0_packet(self, stream_definition_id, length):
...
-                                                          'PTCB0' : 1,
-                                                          'PTCB1' : 1,
-                                                          'PTCB2' : 1,
-                                                          'PA0' : 1,
-                                                          'PA1' : 1,
-                                                          'PA2' : 1}
-
-        config.process.calibration_coeffs['cond_calibration_coeffs'] = {'g' : 1, 'h' : 1, 'I'  :1, 'j' : 1,
-                                                         'CTcor' : 0.1, 'CPcor' : 0.2}
+        config =\
```
lukecampbell, line 121: Can we make config a `DotDict` here?

``` python
from pyon.utils.containers import DotDict
config = DotDict()
config.process = ...
```

ion/processes/data/transforms/ctdbp/test/test_ctdbp_L1.py
```diff
@@ -295,7 +306,7 @@ def _check_granule_from_transform(self, granule):
         rdt = RecordDictionaryTool.load_from_granule(granule)
 
         self.assertTrue(rdt.__contains__('pressure') and rdt.__contains__('temp') and rdt.__contains__('conductivity'))
-        self.assertTrue(rdt.__contains__('time') and rdt.__contains__('lat') and rdt.__contains__('lon'))
+        self.assertTrue(rdt.__contains__('time'))
```
lukecampbell, line 309: I recommend using `assertIn` the message is friendlier and you're using the proper methods.  If we change rdt to override `__iter__` this test will fail because `__contains__` will be different.

``` python
self.assertIn('time', rdt)
```

ion/processes/data/transforms/ctdbp/test/test_ctdbp_chain_L0_L1_L2.py
```diff
@@ -316,6 +316,8 @@ def _launch_transform(self, name_of_transform = '', data_proc_def_id = None, inp
         config = None
         if name_of_transform == 'L1':
             config = self._create_calibration_coefficients_dict()
+        elif name_of_transform == 'L2_density':
+            config = {'lat' : 32.7153, 'lon' : 117.1564}
```
lukecampbell, line 320: Because the configuration is localized to the specific process can we place it in `config.process`?

``` python
from pyon.utils.containers import DotDict
config = DotDict()
config.process.lat = 32.7153
config.process.lon = 117.1564
```

ion/processes/data/transforms/ctdbp/ctdbp_L2_density.py
newbrough, line 36: Luke,

I'm not sure I would agree with using get_safe here -- if these are
supposed to be required values, then we want it to fail with an error if
they are skipped.  Having a default value like this to fall back to
silently hides the problem.  Lat and lon are not just markers of where the
data is coming from, they are involved in the calculation directly.

Jonathan

lukecampbell, line 36: That's absolutely acceptable and encouraged to raise an error when an error case happens, I recommend that the error is descriptive and appropriate though. Here it would have been an `AttributeError` which would actually crash the entire process since it is not an `IonException` and the message would have simply been `lat`. 
You could try:

``` python
lat = self.CFG.get_safe('lat',None)
if lat is None: raise BadRequest('latitude is required to compute...')
```

PR #695, opened 2013-03-14
CIDEVSA-500: use new RR.retire( ) in all SA resources, and use better unit testing

ianfixes:
This set of commits aligns SA's resource CRUD methods with the new RR.retire( ) functionality.  The services' unit tests have also been upgraded, so now the following checks happen on all CRUD operations (and are unit-testable to verify so):
- Don't allow resource creation/update without a name
- Don't allow resource creation of a duplicate name
- Don't allow resource updating to a duplicate name
- Don't allow resource creation with the _id field set
- Require the _id field to be set in update operations
- Don't allow a create/read/update/delete/force_delete method to accept a resource of the wrong type

UNIT and SMOKE tests pass.

Suggest moving ion/services/sa/test/helpers.py to ion/util in the future.


mmeisinger:
I wonder where the name must be unique idea comes from. I don't think this is true. It's desirable that names are expressive and unique but not mandatory from a system or user experience point of view. There may be good reasons why names are the same, e.g. similar thing in different Orgs, a replacement, human choice.

Also it currently thwarts preloading OOI resources, where certain names appear twice.

I think we may want to relax this requirement for some SA resource types but this can be discussed.

ianfixes:
I have no preference.  It was enforced in enough places that I centralized it.  
- All IMS/ObsMS resources https://github.com/ooici/coi-services/blob/26b63da084a3319916dcf72a2732ba004735096b/ion/services/sa/resource_impl/resource_impl.py#L303
- Transforms https://github.com/ooici/coi-services/pull/695/files#L9L115
- Data Products https://github.com/ooici/coi-services/pull/695/files#L11L44

Here are some other examples:
- https://github.com/ooici/coi-services/blob/master/ion/services/dm/presentation/catalog_management_service.py#L31
- https://github.com/ooici/coi-services/blob/master/ion/services/dm/inventory/index_management_service.py#L74
- https://github.com/ooici/coi-services/blob/master/ion/services/dm/distribution/pubsub_management_service.py#L46
- https://github.com/ooici/coi-services/blob/master/ion/services/dm/inventory/index_management_service.py#L161

It might be a wise idea to define an "is distinct from" operator for each resource, and then we could handle all the creates/updates in identical fashion.  (The default would be a._id != b._id, for most SA resources it would be a.name != b.name, etc)

PR #697, opened 2013-03-14
added data dor dds requests to fill index value

mbarry02:


lukecampbell:
We can't have relaodNMinutes set to 1

PR #698, opened 2013-03-14
add get data for dds response

mbarry02:
add get data for dds response


lukecampbell:
Must have missed this...

PR #705, opened 2013-03-14
Updates connection_id container for granule

lukecampbell:
Updates Ion-defs to include granule container connection info


blazetopher:
:+1:

PR #728, opened 2013-03-19
Adds binding support to data process 2

lukecampbell:
This finishes the full reverse compatbility and provides a test to
verify that data processes launched with explicit data process
definitions maintain the binding support needed for reverse
compatibility.


blazetopher:
:+1:

PR #733, opened 2013-03-20
Rdt dims

lukecampbell:


blazetopher:
Merges multi-dimensional brick slicing utilities  :+1:

PR #746, opened 2013-03-21
forgot to add related unit tests -- no changes in code

carueda:
add unit tests for platform attribute grouping according to similar 
           monitoring rate, and aggregation of attribute values for purposes of creation
           of granule

### DIFF
--- a/ion/agents/platform/test/test_platform_resource_monitor.py
+++ b/ion/agents/platform/test/test_platform_resource_monitor.py
@@ -1,0 +1,168 @@
+#!/usr/bin/env python
+
+"""
+@package ion.agents.platform.test.test_platform_resource_monitor
+@file    ion/agents/platform/test/test_platform_resource_monitor.py
+@author  Carlos Rueda
+@brief   Unit test cases related with platform resource monitoring
+"""
+
+__author__ = 'Carlos Rueda'
+__license__ = 'Apache 2.0'
+
+#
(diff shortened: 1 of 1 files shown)


ion/agents/platform/test/test_platform_resource_monitor.py
```diff
@@ -0,0 +1,168 @@
...
+            self.assertIn(k, groups)
+            attr_ids = set(d['attr_id'] for d in groups[k])
+            self.assertEquals(set(expected_groups[k]), attr_ids)
+
+    def test_attr_grouping_by_similar_rate(self):
+
+        self._verify_attr_grouping(
+            platform_id="LJ01D",
+            expected_groups={
+                2.5 : ["input_voltage"],
```
MauriceManning, line 79: How are these grouped? What is the publish interval for 2.5/4.0/5.0 poll rates?

carueda, line 79: See https://github.com/ooici/coi-services/blob/master/ion/agents/platform/platform_resource_monitor.py#L65 for how the attrs are grouped: essentially, it groups the attrs by "similar" poll rate so a single greenlet is launched for each group. "similarity" is given by a group size in seconds, so attributes having poll rates differing by at most that group size will go to the same group.

The publish rate is set to be the minimum of the polling rates.

PR #753, opened 2013-03-22
Rename Parameters with their ID to prevent duplicate parameters inside data containers

lukecampbell:
Data Containers such as the coverage or record dictionary tool or parameter dictionary will fail with parameters with the same name inside the data containers. We designed the framework to allow multiple parameters with the same name in the system but forbid it inside the data containers under the premise that we would never have a need for two things labeled the same. 

A specific use case for this arose when discussing L1/L0 parameters where the L1 parameters have identical names but different units from their L0 counterparts. We are now conquering this issue by appending the parameter identifier to the parameter name. 

Backwards compatibility is maintained by giving the data containers the ability to identify field names with the prefix of the new parameter names.  What was once `time` is now `time_PD7` but the RDT will still recognize it as `time` in cases where there is not more than one. If there are two parameters with the same name, explicit field names should be used:

``` python
rdt['salinity_PD3'] = ...
rdt['salinity_PD4'] = ...
```

### DIFF
--- a/ion/agents/instrument/instrument_agent.py
+++ b/ion/agents/instrument/instrument_agent.py
@@ -862,7 +862,11 @@ class InstrumentAgent(ResourceAgent):
                             data_arrays['time'][i] = tv    
             
             for (k,v) in data_arrays.iteritems():
-                rdt[k] = numpy.array(v)
+                try:
+                    rdt[k] = v
+                except:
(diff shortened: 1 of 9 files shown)


lukecampbell:
Not Ready Yet, still a lot of problems

PR #763, opened 2013-03-24
Alertfix

MauriceManning:
fix a couple items where indexing into struct vs accessing the class and fix index to get stream_name


mmeisinger:
OBE

PR #772, opened 2013-03-25
Updated GeoUtils to calculate midpoint

seman:


ion/util/geo_utils.py
```diff
@@ -1,12 +1,79 @@
...
+            geo_index_obj.lat, geo_index_obj.lon = GeoUtils.midpoint_shortest( north_lat=north_lat, west_lon=west_lon, south_lat=south_lat, east_lon=east_lon)
+        elif distance == GeoUtils.DISTANCE_LONGEST:
+            geo_index_obj.lat,  geo_index_obj.lon = GeoUtils.midpoint_longest( north_lat=north_lat, west_lon=west_lon, south_lat=south_lat, east_lon=east_lon)
+        else:
+            raise BadRequest("Distance type not specified")
         return geo_index_obj
+
+    @staticmethod
+    def midpoint_shortest(north_lat, west_lon, south_lat, east_lon):
+        g = Geod(ellps='clrk66')
```
blazetopher, line 54: I noted this in the original script I attached, but never mentioned it in any of the emails.  We should discuss this with a subject matter expert (SME) to determine which ellipsoid we should be using for these calculations - there are a large number to chose from.  We may also want to make this configurable (an argument) and then make the default whatever recommendation the SME provides.

The same goes for the midpoint_longest calculation.

ion/util/geo_utils.py
```diff
@@ -1,12 +1,79 @@
...
+        #        0         1            111   km
+        #        1         0.1          11.1  km
+        #        2         0.01         1.11  km
+        #        3         0.001        111   m
+        #        4         0.0001       11.1  m
+        #        5         0.00001      1.11  m
+        #        6         0.000001     0.111 m
+        #        7         0.0000001    1.11  cm
+        #        8         0.00000001   1.11  mm
+        rlon = "%.6f" % rlon
```
blazetopher: Is there a particular reason you're turning these values into strings?

ion/util/test/test_geo_utils.py
```diff
@@ -0,0 +1,212 @@
...
+
+        # Basic test
+        #TODO. do we really want calc_geospatial_point_center() to return string?
+        mid_point = GeoUtils.calc_geospatial_point_center(geospatial_bounds)
+        lat = float(mid_point.lat)
+        lon = float(mid_point.lon)
+        self.assertAlmostEqual(lat, 0.0)
+        self.assertAlmostEqual(lon, 0.0)
+
+        mid_point = GeoUtils.calc_geospatial_point_center(geospatial_bounds)
```
blazetopher: Is this the same as lines 30 - 34?

PR #776, opened 2013-03-26
Removes some log debugs that are not required

swarbhanu:


MauriceManning:
Please cmmt out versus remove so that the debug could be re-activated for detailed debugging if required.

swarbhanu:
pushed in a commit. log debugs are now commented and not removed.

PR #780, opened 2013-03-26
add aggregate status calc to instrument agent

MauriceManning:


edwardhunter:
2 things in instrument_agent:
1. You don't need to use get_status(), that is just a way to turn the object back to a dict for get alarms functions. You can just reference the objects directly.
2. The logic in process aggregates looks wrong to me (I didn't study it super carefully). It looks like it doesn't actually aggregate anything but just updates such that the last value with a particular aggregate status type will overwrite the others and that is all that is reflected. I think you want a slightly different loop logic, unless I'm missing something. Check it out and let me know.

PR #782, opened 2013-03-26
Added ability to cache predicates for in-memory lookup inside EnhancedResourceRegistryClient

ianfixes:
This commit addresses the timeout error present in some extensive "launch platform agent" tests, caused by excessive calls to RR.find_subjects and RR.find_objects.

This has been fixed by adding capability the EnhancedResourceRegistryClient.

old way:

``` python
# makes 2 calls to resource registry
RR2 = EnhancedResourceRegistryClient(self.clients.resource_registry)
models1 = self.RR2.find_objects("device1", PRED.hasModel, RT.InstrumentModel, True)
models2 = self.RR2.find_objects("device2", PRED.hasModel, RT.InstrumentModel, True)
```

new way:

``` python
# makes 1 call to resource registry
RR2 = EnhancedResourceRegistryClient(self.clients.resource_registry)
RR2.cache_predicate(PRED.hasModel)
models1 = self.RR2.find_objects("device1", PRED.hasModel, RT.InstrumentModel, True)
models2 = self.RR2.find_objects("device2", PRED.hasModel, RT.InstrumentModel, True)
```

Note that no change is required to existing find_\* calls.

### DIFF
--- a/ion/services/sa/instrument/agent_configuration_builder.py
+++ b/ion/services/sa/instrument/agent_configuration_builder.py
@@ -37,11 +37,34 @@ class AgentConfigurationBuilder(object):
 
     def __init__(self, clients):
         self.clients = clients
-        self.RR2 = EnhancedResourceRegistryClient(self.clients.resource_registry)
+
+        if not hasattr(self.clients, "RR2") or {} == self.clients.RR2:
+            self.clients.RR2 = EnhancedResourceRegistryClient(self.clients.resource_registry)
+            self._update_cached_predicates()
+
+        self.RR2 = self.clients.RR2
+
+        if not isinstance(self.RR2, EnhancedResourceRegistryClient):
+            raise AssertionError("Type of self.RR2 is %s not %s" %
+                                 (type(self.RR2), type(EnhancedResourceRegistryClient)))
+
         self.agent_instance_obj = None
         self.associated_objects = None
         self.last_id            = None
         self.will_launch        = False
+
+
+    def _update_cached_predicates(self):
+        # cache some predicates for in-memory lookups
+        for pred in [PRED.hasOutputProduct,
+                     PRED.hasStream,
+                     PRED.hasStreamDefinition,
+                     PRED.hasAgentInstance,
+                     PRED.hasAgentDefinition,
+                     PRED.hasDataset,
+                     PRED.hasDevice]:
+            self.clients.RR2.cache_predicate(pred)
+
 
     def _lookup_means(self):
         """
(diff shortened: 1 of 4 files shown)


mmeisinger:
Caching is a good idea but let's just be very careful. There could be very confusing bugs originate from this, especially if the cache is never refreshed/cleared or different service workers use different states of the cache.

ianfixes:
That's correct.  In fact, the changes like this one address that exact problem:
https://github.com/ooici/coi-services/pull/782/files#L1R628

The problem is mainly felt in tests.  Normally, you would be caching when you needed to do a lot of read-only stuff, not when you're doing a mix of read/write.

In that sense, the problem is no different than doing a cache operation manually.  This just centralizes it and makes it possible for all the calls to benefit from the cache.

shenrie:
I am curious why this caching is needed if the get extended is only doing one call to the RR for finds anyway?

---

System Architect

ianfixes:
"Caching" may be a strong word for what's happening here.  In the past, Michael had asked that instead of doing many separate RR.find_subj/obj calls, I just get the entire list of associations in one call and do the rest in memory.

This is that same pattern, being used on a lower level.  

It is still a manual process, and is only relevant for cases where you expect to read the same predicate very often (like building a tree of hasDevice, hasSite, etc).  It is not meant to be a cache across calls, just a way to make many find functions work from a single RR op in a single service method call.

Is the problem that the cached associations should be explicitly disposed of after use?  

This is addressing a specific problem in the AgentConfgurationBuilder, because some complex platform launches seem to be taking several minutes to build a configuration.

shenrie:
That is a good pattern to follow. I was just looking to understand what you are doing here.

I would not cache the associations unless you have a way to invalidate that cache if those cached objects change. You can see a patten for this in the service gateway, where I keep a cache of user data based off of the set of roles associated with a user. However, if a role changes for a user then that user's info is removed from this cache to that any changes will not be kept stale in this cache. This is done by listening for specific ResourceModified events which trigger a UserRoleCacheResetEvent which in turns invalidate this cached data associated with the user in question.

---

System Architect

ianfixes:
I think it would be more appropriate to invalidate the cache every time a new service method call is done (every time get_context changes).  What's the best way to do this?

shenrie:
That's fine…whatever works best. You should probably ask Dave Foster about the get context changing.

---

System Architect

mmeisinger:
If you are dependent on invalidating when a service operation is called, then it will not work in a case where you have 2 worker processes. You really need to think and test for this case, which makes caching database state almost impossible.

Requests go round robin to each worker and if both keep a cache then one may be out of sync if an association was added.

I do understand the pattern of reading the associations once and then just answering the questions within the same request from this "cache", but this is a very limited use case for operations that do lots of finds in a row and then exit. The cache must not live beyond a service operation.

ianfixes:
> The cache must not live beyond a service operation.
> 
> If you are dependent on invalidating when a service operation is called, then it will not work in a case where you have 2 worker processes.

Agree.  For now, these cases can be avoided by creating a new instance of the EnhancedResourceRegistryClient where appropriate.  For instance, I do this within the AgentConfigBuilder instead of within the service.

PR #786, opened 2013-03-27
Platform instrument hierarchy launch

carueda:
Set-up of the launch tests now enabled to create and associate instruments to desired platforms in the network. The concrete test is:

```
bin/nosetests -sv ion/services/sa/observatory/test/test_platform_launch.py:TestPlatformLaunch.test_platform_hierarchy_with_some_instruments
```

which exercises the 2 currently available SBE37 simulators in a multi-level network comprising 13 platforms.

NOTE: this test is skipped to avoid potential timeout issues on the buildbots, but it runs fine with a reasonable endpoint.receive.timeout (which I'm setting to 180 in my pyon.local.yml).  

But please review, test, and merge at the earliest.


ianfixes:
These mods to enhanced RR client and agent_config_builder are OK.

PR #803, opened 2013-03-27
bump sbe driver version

wfrench:
Bumped the driver version used in tests to include the latest connection recovery code and bug fixes.

Passed smoke tests and all tests in test_instrument_agent.py


carueda:
By searching "seabird_sbe37smb_ooicore" in coi-services I see 18 usages (with a variety of concrete egg versions btw).  Perhaps a parameter in pyon.yml could be added to point to the latest and greatest so we can use it and then inherit the value without further changes upon this kind of updates?  Just a suggestion.  I will now update my own usages.

PR #809, opened 2013-03-28
workarounds for the recent issues related with the new hasNetworkParent association

carueda:
, which is not yet handled by the enhanced registry. I just
explicitly  implemented unresolved functions.

Locally tested with the platform agent tests and the failed test in TestInstrumentManagementServiceIntegration


carueda:
Not used, and not needed any more.

PR #811, opened 2013-03-28
End to end working implementation of lookup tables.

lukecampbell:
Requires [Ion-defs PR 321](https://github.com/ooici/ion-definitions/pull/321)


blazetopher:
:) :+1:

PR #814, opened 2013-03-28
Removes "guessing" behavior from enhanced RR client.

ianfixes:
Tests are passing.


mmeisinger:
Thanks. If if all works, merge is fine

PR #823, opened 2013-03-30
fix service gateway inability to handle dicts, add unicode test

pkediyal:
created a new branch that works around having to deal with problems associated with merging my pull request. I will close the other one.


pkediyal:
created a new branch that works around having to deal with problems associated with merging my pull request. I will close the other one.

PR #828, opened 2013-04-01
more instrument sims included and preliminary port agent coordination

carueda:
- add the 5 recently enabled instrument simulator instances; the 7 instruments exercised in test_platform_hierarchy_with_some_instruments but only 2 still enabled due to issues under investigation; 
- include preliminary basis for coordination with port agents (new PortAgentHelper and copy of PortAgentPacket from mi.core.instrument.port_agent_client)
- per discussion, no port agent coordination is needed after all (at least for the iteration).


carueda:
Will submit a new PR with more updates and alignment, including adjustments upon recent updates by Edward in base classes.

PR #830, opened 2013-04-01
plumb timeout into agent tests that access using service gateway

newbrough:
altered ResourceAgentViaServiceGateway to support timeout in HTTP service calls in same way service_api.py invokes calls from ion-ux.

### DIFF
--- a/ion/agents/instrument/test/test_gateway_to_instrument_agent.py
+++ b/ion/agents/instrument/test/test_gateway_to_instrument_agent.py
@@ -117,8 +117,8 @@ class ResourceAgentViaServiceGateway(ResourceAgentClient):
         return _process_gateway_request(resource_id, "get_capabilities", agent_get_capabilities_request, requester)
         
     # XXX            
-    def gw_execute_agent(self, resource_id="", command=None, requester=None):
-        return self._gw_execute( 'execute_agent', resource_id, command, requester)
+    def gw_execute_agent(self, resource_id="", command=None, requester=None, timeout=300):
+        return self._gw_execute( 'execute_agent', resource_id, command, requester, timeout)
     
     # XXX    
     def gw_get_agent(self, resource_id='', params=[], requester=None):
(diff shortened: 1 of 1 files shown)


ion/agents/instrument/test/test_gateway_to_instrument_agent.py
```diff
@@ -151,15 +151,16 @@ def gw_ping_resource(self, resource_id='', requester=None):
...
-    def _gw_execute(self, op, resource_id, cmd, requester=None):
+    def _gw_execute(self, op, resource_id, cmd, requester=None, timeout=300):
 
         agent_cmd_params = IonObjectSerializer().serialize(cmd)
 
         agent_execute_request = {  "agentRequest": {
             "agentId": resource_id,
             "agentOp": op,
-            "expiry": 0,
+#            "expiry": 0,
```
shenrie, line 161: Why did you comment out this? A zero indicates to the framework that the credentials used by the request should not expire. I would think that you would want this for testing purposes.

newbrough, line 161: just duplicated what i saw in service_api.py so the calls would be
identical.  i thought this was maybe the old-style of timeout, but i guess
not.  you want this uncommented?

shenrie, line 161: Well, now I am concerned that this was commented out in the service_api…that's no bueno. These are two different things. The expiry has to do with the how long the credentials are good for ( related to the requestor value ) and the timeout is specific to that request. It should not be commented out in either place.

---

System Architect

newbrough, line 161: does the 0 value means the creds never time out or that they immediately
timeout (this request only)?

shenrie, line 161: Never time out.

---

System Architect

PR #834, opened 2013-04-02
typo fixed / don't clear the cache here.

ianfixes:


ianfixes:
merged into carueda's personal fork

PR #838, opened 2013-04-02
Adds cleanup for parameter dictionary test

tgiguere:


lukecampbell:
You don't need to delete it, it's part of the container for the lifetime of the system
It's part of preload

PR #839, opened 2013-04-02
Fix memory leak related to stream associated with data product that is left around

swarbhanu:
- Cleans up stream associated with data product in delete_data_product.
- Adds test code to verify that the stream is indeed cleaned up when delete_data_product is used.

### DIFF
--- a/ion/util/enhanced_resource_registry_client.py
+++ b/ion/util/enhanced_resource_registry_client.py
@@ -246,6 +246,17 @@ class EnhancedResourceRegistryClient(object):
 
         return
 
+    def assign_stream_definition_to_data_product_with_has_stream_definition(self, stream_definition_id='', data_product_id=''):
+
+        self.RR.create_association(subject=data_product_id, predicate=PRED.hasStream, obj=stream_definition_id)
+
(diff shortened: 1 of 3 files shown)


ion/util/enhanced_resource_registry_client.py
```diff
@@ -246,6 +246,17 @@ def retire(self, resource_id='', specific_type=None):
...
+
+        self.RR.create_association(subject=data_product_id, predicate=PRED.hasStreamDefinition, object=stream_definition_id)
+
+    def assign_stream_to_data_product_with_has_stream(self, stream_id ='', data_product_id=''):
+
+        self.RR.create_association(subject=data_product_id, predicate=PRED.hasStream, object=stream_id)
+
+    def assign_dataset_to_data_product_with_has_dataset(self, dataset_id = '', data_product_id = ''):
+
+        self.RR.create_association(subject=data_product_id, predicate=PRED.hasDataset, object=dataset_id)
```
lukecampbell: I think these already exist dynamically; from what I understand about the enhanced resource registry.

swarbhanu:
yes, right. reverted that bit. Thanks.

PR #841, opened 2013-04-02
R2 full tests

lukecampbell:
This PR adds test and test utilities to verify functionality for DM capabilities in a distributed system.

The largest challenge is resource collection in between tests. Before we were relying on preload, which is also how the system is actually bootstrapped.  We can't rely on preload for the R2-Full integration tests because subsequent preloads will crash (duplicates) and nothing cleans up the resources between tests.


blazetopher:
:+1:

PR #846, opened 2013-04-02
Fixes Lookup Tables: Adds static lookup keys

lukecampbell:


lukecampbell:
Success is just around the corner...

PR #854, opened 2013-04-03
Stored value transform

lukecampbell:


blazetopher:
It's all coming together!! :facepunch:

PR #856, opened 2013-04-03
Attachment Reference Support

lukecampbell:
When `assign_data_product` is called in `Data Acquisition` all attachment documents labeled as a referenced are parsed and then ingested into the object store.


ion/services/sa/acquisition/data_acquisition_management_service.py
```diff
@@ -228,6 +229,11 @@ def assign_data_product(self, input_resource_id='', data_product_id=''):
         data_producer_obj = DataProducer(name=data_product_obj.name, description=data_product_obj.description)
         data_producer_id, rev = self.clients.resource_registry.create(data_producer_obj)
         log.debug("DAMS:assign_data_product: data_producer_id %s" % str(data_producer_id))
+        for attachment in self.clients.resource_registry.find_attachments(data_product_id, include_content=False, id_only=False):
+            if attachment.attachment_type == AttachmentType.REFERENCE:
+                parser_id = attachment.context.parser_id
+                if parser_id:
+                    self.register_producer_qc_reference(data_producer_id, parser_id, attachment._id)
```
lukecampbell, line 236: in `assign_data_product`: for each Attached document, parse the document and update the producer's context with the document keys.

PR #874, opened 2013-04-05
CIDEVSA-489 and OOIION-791 - deployment activation

ianfixes:
OK TO MERGE.  All SMOKE, UNIT, and sa-related INT tests pass.

This code replaces (nearly completely) the functionality of deployment activation.  
- Site data products have been completely eliminated.  They are still in the service spec but will be removed in a separate PR (combining ion-defs) to keep it manageable.
- Matching devices to sites in a deployment by CSP solving is now only done as a fallback.
- "reference_designator"/PlatformPort is now the preferred method of matching sites.
- Different deployment contexts are supported, with different behavior concerning how child sites/devices are treated.  Are they allowed?  Are they included?  Depends on context.

Old test cases have been updated (for backwards compatibility, use `context=IonObject(OT.CabledNodeDeploymentContext)` when constructing deployment objects).

New test cases create a hierarchy of 1 platform with 3 subplatforms, each subplatform having 3 instruments.  2 devices/sites in each set of 3 have an identical model, which makes the old method of model-based matching impossible.  The tests demonstrate that this structure can still be deployed using the `port_assignments` field appropriately.


ianfixes:
Safe to merge.

PR #882, opened 2013-04-05
Removes OBE test

lukecampbell:
The test originally tested the transforms ability to compute parameter
functions. This functionality is tested in 8+ other tests in data
process management integration tests. The test failed because the
underlying parameters were hardcoded and changed. The new pattern in
some of my tests is to use parameter helper but rewriting the test is
mute since it's already tested elsewhere.


blazetopher:
:+1:

PR #889, opened 2013-04-05
Tests alerts coming from a sim that delays data

swarbhanu:
- Integration tests for two types of alerts set up works.


swarbhanu:
Made a pull request for the above comment here:
https://github.com/ooici/coi-services/pull/893

Please merge it if it is okay. The comms status is now being calculated accurately. Test passes.

PR #893, opened 2013-04-05
Test to  show that the late_alert causes the rollup comms status to be set to warning when the extended instrument resource is retrieved

swarbhanu:
- Now the comms status roll up works. 
- added assertion in integration test for the extended instrument's computed attribute to check that it works
- removed the DeviceCommsEvent from aggregate status computation code. Will remove more instances of DeviceCommsEvent soon.

### DIFF
--- a/ion/services/sa/test/test_instrument_alerts.py
+++ b/ion/services/sa/test/test_instrument_alerts.py
@@ -16,7 +16,7 @@ from pyon.agent.agent import ResourceAgentClient
 from pyon.util.context import LocalContextMixin
 from pyon.event.event import EventSubscriber
 
-from interface.objects import StreamAlertType, AggregateStatusType
+from interface.objects import StreamAlertType, AggregateStatusType, StatusType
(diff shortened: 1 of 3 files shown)


ion/services/sa/test/test_instrument_alerts.py
```diff
@@ -369,5 +369,7 @@ def callback_for_alert(event, *args, **kwargs):
             elif c.name == 'late_data_warning':
                 self.assertEqual(c.message, 'Expected data has not arrived.')
 
+        extended_instrument = self.imsclient.get_instrument_device_extension(instrument_device_id=instDevice_id)
+        self.assertEqual(StatusType.STATUS_WARNING, extended_instrument.computed.communications_status_roll_up.value)
```
swarbhanu, line 373: Here, we check the communication status. This pull request is open. Please merge it if it is okay.

mmeisinger:
OBE

PR #904, opened 2013-04-09
Fixes failing pycc test in buildbot 

swarbhanu:
- Extended data process definition and data process attributes were not being filled. 
- Also, there is an association between data process definition and data process (PRED.hasDataProcess) which was not being created in create_data_process() when data process definition is provided.

### DIFF
--- a/ion/services/sa/process/data_process_management_service.py
+++ b/ion/services/sa/process/data_process_management_service.py
@@ -302,6 +302,8 @@ class DataProcessManagementService(BaseDataProcessManagementService):
 
         self.clients.resource_registry.create_association(subject=dproc_id, predicate=PRED.hasProcess, object=pid)
 
+        if data_process_definition_id:
+            self.clients.resource_registry.create_association(data_process_definition_id, PRED.hasDataProcess ,dproc_id)
 
         return dproc_id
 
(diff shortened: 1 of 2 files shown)


ion/services/sa/process/data_process_management_service.py
```diff
@@ -715,6 +717,8 @@ def get_data_process_definition_extension(self, data_process_definition_id='', e
                 if hasattr(att, 'content'):
                     delattr(att, 'content')
 
+        data_process_ids, _ = self.clients.resource_registry.find_objects(data_process_definition_id, PRED.hasDataProcess, RT.DataProcess, True)
```
MauriceManning, line 720: why doesnt the decorator work here? #@hasProcessDefinition on the extended resource def? Same for in data products below?

ion/services/sa/process/data_process_management_service.py
```diff
@@ -742,5 +746,8 @@ def get_data_process_extension(self, data_process_id='', ext_associations=None,
                 if hasattr(att, 'content'):
                     delattr(att, 'content')
 
+        input_data_products, _ = self.clients.resource_registry.find_objects(data_process_id, PRED.hasInputProduct, RT.DataProduct, True)
```
MauriceManning, line 749: https://github.com/ooici/ion-definitions/blob/master/objects/data/sa/xsaresources.yml#L38

swarbhanu:
will need to find why the extended data process definition attribute is not getting filled in this test through the usual decorator framework. closed pull request until then.

PR #911, opened 2013-04-09
aggregation status now handled using a "push" approach

carueda:
- aggregation status now handled using a "push" approach: all calculations and subsequent publication of derived events are triggered by events received from children agents. 
- Using new supporting class StatusUtil. Specific aparams handled are aparam_child_agg_status and aparam_aggstatus.  
- test_platform_status exercises this updates with various transitions.

Note that this is about updates related with aparam_child_agg_status and aparam_aggstatus, but nothing is actually done with aparam_rollup_status yet. I have used observatory_util.py as a reference to make sense of the various status concepts (aggregate, consolidated, rollup), and I think this PR should cover most of what it is called "rollup status." In general, the new push based mechanism in this PR is the preferable approach to propagate status updates in the platform hierarchy.


MauriceManning:
very nice

PR #914, opened 2013-04-10
Fixes coi_pycc tests

lukecampbell:
XN and XP cleanup crashes subsequent tests and possibly exchange
management, I removed it in favor of using addCleanup and letting the
exchange management to cleanup resources appropriately in lieu of
explicit cleanup which violates the workflow (and crashes).


daf:
:bird:

PR #919, opened 2013-04-10
Add route to SGS to delete attachments

daf:


daf:
Merging, will fix return value pattern in a subsequent PR.

shenrie:
One other thought I had about this is that other than supporting the restful DELETE operation, this route is really unnecessary as the same thing can be done by posting to /ion-services/resource_registry/delete_attachment with the attachment_id parameter. The create and get attachments needed to be different as they deal with the binary content transfer, but the delete does not, so it could be a standard service operation.

PR #926, opened 2013-04-11
Preload external data

newbrough:
support external dataset objects in preload


mmeisinger:
Please use the self._basic_resource_create call for most preload resource creates. It does all the steps for you and has plenty of arguments to supply additional and override arguments to the objects

mmeisinger:
Looks good on quick glance

PR #931, opened 2013-04-12
Use normal serviceRequest format for SGS create attachment

daf:


shenrie:
Looks much better...thanks!

PR #940, opened 2013-04-15
Adds skip test on version checking for h5stat

lukecampbell:


daf:
Wary about having this sort of thing at the module level, automatic subprocess calls on import etc.  Any way to put this inline with tests instead?  Still merging for now.

PR #963, opened 2013-04-18
Fix missed rename of method for IMS _set_device_aggregate_status

daf:


daf:
Exactly the same as #962, nice.

PR #966, opened 2013-04-18
Parameters

lukecampbell:
First pass at Dynamic QC Parameter creation and utilization


lukecampbell:
Need to get the ball rolling on this, tests still fail but at least some stuff works.

PR #982, opened 2013-04-23
Preload external data

newbrough:
new parser to handle binary SBE52 data files and multiple profiles within a single file
preload tabs to handle external dataset-related resources
test class that preloads resources, starts dataset agent to read file, checks resulting granule values


MauriceManning:
very nice.

PR #997, opened 2013-04-25
Use managed endpoint API

daf:
Updates services/processes/transforms to use the managed endpoint API so that endpoint failures are noticed by the owning process.  Should lead to better accountability for issues, especially in transforms.


jamie-cyber1:
:ok_woman:

ion/core/process/transform.py
```diff
@@ -31,16 +31,9 @@ def on_start(self):
...
-        self._listener = gevent.spawn(self._rpc_server.listen)
-
-    def on_quit(self):
-        '''
-        Closes the listener.
-        '''
-        self._rpc_server.close()
-        self._listener.join(5)
-        super(TransformBase, self).on_quit()
+        self._rpc_server = self.container.proc_manager._create_listening_endpoint(from_name=self.id,
```
lukecampbell, line 34: I don't even think we use this at all.

PR #1004, opened 2013-04-26
add launch processing state to platform agent

MauriceManning:
Add block state while platform launches children.


carueda:
Looks good. Just wondering about the other various "busy" states happening during the processing of requests like INITIALIZE, GO_ACTIVE, RUN, etc, (with propagation to children); is it the idea to also introduce corresponding _ING_ (eg., GOING_ACTIVE) states?  Or perhaps a single BUSY state would be enough?

PR #1013, opened 2013-04-29
Platform alerts

MauriceManning:
add alerts to platform agent instance and add pubrate to instrument, extDatasett and platform agent instance agent configs


MauriceManning:
Suggest a column name and I will change it tonight.

PR #1022, opened 2013-04-30
Fix various attributes of platform device extensions and site extensions

ianfixes:


ianfixes:
SMOKE tests passing, governance tests passing.  New tests added.

PR #1040, opened 2013-05-06
Fixes

lukecampbell:
Adds RDT support for accepting stream definition objects
Tightens up the RDT performance for initializing from a granule
Fixes various interceptor errors


blazetopher:
:+1:

PR #1047, opened 2013-05-07
Removes QC parameters from UI

lukecampbell:
Only Global Range is support in real time, all other QC parameters are removed from UI processing.


blazetopher:
:+1:

PR #1048, opened 2013-05-07
Add NeedLogin field to SGS error response on cred expiry

daf:


ion/services/coi/service_gateway_service.py
```diff
@@ -364,9 +365,13 @@ def build_error_response(e):
     result = {
         GATEWAY_ERROR_EXCEPTION : exec_name,
         GATEWAY_ERROR_MESSAGE : str(e.message),
-        GATEWAY_ERROR_TRACE : full_error
+        GATEWAY_ERROR_TRACE : full_error,
+        GATEWAY_ERROR_NEEDLOGIN : False
     }
 
+    if isinstance(e, Unauthorized) and 'expired' in str(e):
+        result[GATEWAY_ERROR_NEEDLOGIN] = True
```
shenrie, line 373: I am not a big fan of this... I think it is letting the domain of the UI bleed into the gateway, which should be agnostic since it will not only be used by the UI. I would rather see the return contain something else and have the UI flask layer interpret it to require a login

daf:
Ok - I guess I can just perform the same check on the UI side then.

PR #1062, opened 2013-05-10
Fix RR Service to use RRServiceWrapper

daf:
Unifying the interface to `create`/`create_attachment`


shenrie:
Michael, are you ok with this? It looks ok to me...

mmeisinger:
yes this is good and the one case I missed before :+1:

PR #1087, opened 2013-05-16
Set multi flag on prepare inst device agent association

daf:
Despite the fact we can't really do these associations properly yet, at least give the hint to the UI so it doesn't try to treat it like a singleton.

### DIFF
--- a/ion/services/sa/instrument/instrument_management_service.py
+++ b/ion/services/sa/instrument/instrument_management_service.py
@@ -1921,6 +1921,9 @@ class InstrumentManagementService(BaseInstrumentManagementService):
             'unassign_sensor_device_from_instrument_device', { "sensor_device_id":  "$(sensor_device_id)",
                                                                 "instrument_device_id":  instrument_device_id })
 
+        # the 'InstrumentAgent' association, correct or not here, needs to be multiple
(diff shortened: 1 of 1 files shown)


shenrie:
You should check with Maurice about this, because I don't think there can be multiple agents for a device. This doesn't seem correct to me.

---

System Architect

daf:
Ok so this is related to the "weird case".  I think I was going off the data that was coming in the `associated_resources` field being multiple.

shenrie:
Yes, it's multiple cause is related to the set of instrument models, so that you can display the appropriate sub-set of agents in select box B based on the model selected in select box A.

---

System Architect

daf:
We're going to have to find a way to encode this logic into the prepare data structures themselves then, because association editing is 100% generic.  I'll have a look and see what I can come up with.

shenrie:
That won't work. It will be too static. The UI needs to be more dynamic, so I am providing the set of into needed to do it. This type of association just needs to be handled special as its own pattern.

---

System Architect

daf:
I mean something like 'this is an indirect association through this other association'.  Might even end up being one field.  That way the front-end could be properly generic and just go off of the rules given to it.

PR #1091, opened 2013-05-17
Fixes [None]*len publishing in InstrumentAgent

lukecampbell:
InstrumentAgent would publish an array of Nones for all variables in the
RDT.  This patch limits the variables being published to only those in
the tomato and the rdt. It avoids the parameter function variables like
'salinity' from being published as all Nones and the GRTs.  When a
client receives a granule with actual content [None] != None it uses the
received value in lieu of calculating it.

This may potentionally fix a bug with gaps (unlikely).


blazetopher:
woot! :+1:

PR #1095, opened 2013-05-17
Fixes the memory leak and segfaults

lukecampbell:
A circular reference in the Coverage Model combined with overriding `__del__` prevented the garbage collection module from freeing the instances of the Coverage Model. This patch removes `__del__` and removes the dependency on RTree.

Note: RTree implements a libspatialindex python wrapper using ctypes.  No type checking can be performed on the pointers being passed into the library and as a consequence there are situations where `dynamic_cast<SpatialIndex::ISpatialIndex>` causes a SEGFAULT.  This library has been removed from the coverage model.


blazetopher:
:bomb:

PR #1107, opened 2013-05-21
Greenlet leaks

newbrough:


ion/processes/data/ingestion/science_granule_ingestion_worker.py
```diff
@@ -100,6 +100,9 @@ def on_start(self): #pragma no cover
         self.start_listener()
 
     def on_quit(self): #pragma no cover
+        self.event_publisher.close()
+        self.qc_publisher.close()
+        self.lookup_monitor.stop()
```
lukecampbell, line 105: We're not publishing with `to_name` so closing the publishers is probably a no-op.

The lookup monitor did need to get stopped.

lukecampbell, line 105: Actually the lookup monitors should be added to the processes listeners, not manually managed by the process (if it dies, the listeners are unreferenced and will remain).

PR #1120, opened 2013-05-24
Fix assoc list in IMS prepare to include unassoc'd AgentInstances

daf:


shenrie:
Did you run the PREP tests on this cause they do not seem to be passing

PR #1122, opened 2013-05-25
adjustments related with removals/changes in the CI-OMS interface spec

carueda:
- remove functionality related with changes in the CI-OMS interface specification (in particular re the removal on the 'event type' concept); and do adjustments while starting testing with the actual RSN OMS endpoint implementation (eg., use of the xml/rpc handlers they are using). Still some adjustments to do.
- systematic "monitorCycleSeconds" -> "monitor_cycle_seconds" renaming to use consistent naming and align with actual RSN OMS implementation. 
- Fix PlatformAgentInstance x driver_config cell for MJ01C platform in the preloaded spreadsheet (was referring to wrong ports apparently after a direct copy-n-paste from the LJ01D case). 
- Make code in resource_monitor more robust upon unexpected
  format in received attribute values.

Note: all platform unit and integrations tests passing against the simulator (as usual), and some now starting to pass against the actual RSN implementation, but adjustments for the latter are still ongoing.


ion/agents/platform/resource_monitor.py
```diff
@@ -111,7 +111,10 @@ def _run(self):
                 slept += incr
 
             if self._active:
-                self._retrieve_attribute_values()
+                try:
+                    self._retrieve_attribute_values()
+                except:
+                    log.exception("exception in _retrieve_attribute_values")
```
MauriceManning, line 117: should the exception be logged?

carueda, line 117: Note that log.exception takes the exception automatically from the current frame -- no need to pass it explicitly.  Isn't that sufficient?

PR #1125, opened 2013-05-28
OOI preload revision

mmeisinger:
This PR provides a slew of smaller and medium changes for preload, including:
- Adding temporal bounds to Deployment
- Adding the link for a DataProduct to its site
- Create StreamDefinition for each instrument and derived DataProduct
- Creating DataProducts based on their correct StreamDefinitions (if existing)
- OOI preload now assuming prior existing resources for Orgs etc
- OOI preload now run within "scripted" preload runs
- Improved way of counting effected rows
- General polishes
- Option to exclude categories from load
- Option to clear certain column values before load
- Option to snapshot RR before preload and revert any changes in case of preload failure if debug=True
- Mapping of preload identifiers to maintain existing scenarios

A consequence is that BETA, ALPHA_SYS and BETA_SYS scenarios are deprecated for OOI preload (but still work).


mmeisinger:
Updated and merged

PR #1142, opened 2013-05-30
fix overly aggressive assert; allow None

ianfixes:


ion/services/sa/instrument/status_builder.py
```diff
@@ -274,7 +274,11 @@ def get_status_of_device(self, device_id):
 
 
     def get_aggregate_status_of_device(self, device_id):
-        assert(type("") == type(device_id))
+        if  device_id is not None and type("") != type(device_id):
```
lukecampbell, line 277: One thing I recommend here is comparing the type against `basestring` which will avoid errors popping up for unicode characters. 

```
if device_id is not None and not isinstance(device_id, basestring):
    ...
```

PR #1144, opened 2013-05-30
Patches create_data_process

lukecampbell:
There were a bunch of failling tests, ultimately the cause comes from
the fact that the data process definitions supplied didn't have any
bindings for named streams.  I re-arranged the logic block so that if
the data process definition doesn't supply any bindings it will attach
the streams for the output data products.


blazetopher:
:+1:

PR #1148, opened 2013-05-31
Fixes test_time_search on buildbot maybe

lukecampbell:
I think the issue is that buildbot runs near midnight and that messes up
the span of dates.  By expanding the span of dates we can hopefully get
past this.


blazetopher:
bizarre...

PR #1156, opened 2013-06-03
Discovery

lukecampbell:
Makes Discovery's `SEARCH_BUFFER_SIZE` configurable and defaults to 1024 \* 1024.

### DIFF
--- a/ion/services/dm/presentation/discovery_service.py
+++ b/ion/services/dm/presentation/discovery_service.py
@@ -26,9 +26,9 @@ import time
 import elasticpy as ep
 import heapq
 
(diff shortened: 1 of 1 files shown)


blazetopher:
:+1:

mmeisinger:
As a general good practice I recommend putting the actual config entry into pyon.yml with it's default. This way it gets apparent what is configurable by reading pyon.yml instead of reading the code.

Not required but good for housekeeping.

lukecampbell:
Yup, I agree.  I bumped ion-defe to include the Changes for pyon.yml

PR #1163, opened 2013-06-04
calculate portal status even when not part of master status dict

newbrough:
prior implementation was extracting status of portal devices from main list of device statuses, but several cases had portal devices that were not in the device list of the site/facility.


mmeisinger:
Careful. This PR seems to have had much more impact than anticipated. Please check that this one did not revert other people's changes

daf:
`git-rebase master` yw

blazetopher:
It seems to have re-committed a number of recent commits (recent ones from me are in there for example)

Jonathan - it looks like you may have merged the most recent master and then created and merged the PR?  This effectively re-commits all the commits you merge into master locally.  Rebase is a better approach (as Dave just commented)...

blazetopher:
Didn't notice the direction of the PR - I get it now...sort of :/

PR #1165, opened 2013-06-04
Fix jira1024

newbrough:
new branch with the changes that were garbled in the earlier branch.

changes collect device status for deployment and portal lists directly,
prior release extracted them as a subset of existing device lists.


ianfixes:
There's a problem in here:
https://github.com/ooici/coi-services/pull/1165/files#L2R114

`out_status` is a dict but `child_agg_status.items()` is a list of tuples.  To properly merge the 2 dicts, you need the original line:
`out_status = dict(out_status.items() + child_agg_status.items())`

Is this causing a test failure somewhere?

PR #1177, opened 2013-06-07
Adds bootstrap plugin to load QC references

lukecampbell:
To load QC References enable the `bootstrap.load_references` flag.


blazetopher:
:+1:

PR #1185, opened 2013-06-10
Reference docs

lukecampbell:
For [OOIION-1087](https://jira.oceanobservatories.org/tasks/browse/OOIION-1087) Reference Document Loading in Preload


mmeisinger:
:rocket:

PR #1187, opened 2013-06-11
DirectCoverageAccess Improvements and Fixes

blazetopher:
Extends / improves DCU functionality and tests

Renames SimpleCSVParser to SimpleDelimitedParser

Adds DCA.clean_up function to enable simplified clean up of open coverages and paused streams

Adds TestSimpleDelimitedParser class with tests for parser functionality.

Minor changes to testmanualupload.yml and testcalcoeff.yml to improve test coverage of SimpleDelimitedParser

Increases coverage of DirectCoverageAccess

Improves warnings when DCA is used outside of context management and tests warnings

### DIFF
--- a/ion/util/direct_coverage_utils.py
+++ b/ion/util/direct_coverage_utils.py
@@ -6,10 +6,6 @@
 @author Christopher Mueller
 @brief Utilities for operating with coverages via a direct connection
 """
-from interface.objects import ProcessDefinition, ProcessStateEnum, DataProduct
-from ion.services.cei.process_dispatcher_service import ProcessStateGate
-
-from pyon.public import log
 
 import os
 import yaml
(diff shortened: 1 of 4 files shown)


ion/util/direct_coverage_utils.py
```diff
@@ -28,6 +24,10 @@
 from coverage_model.recovery import CoverageDoctor
 
 
+def warn_user(msg):
+    print msg
```
lukecampbell: If it's a warning can you please replace this with either a logging output or write to sys.stderr

``` python
import sys
sys.stderr.write(msg + '\n')
```

ion/util/test/test_direct_coverage_utils.py
```diff
@@ -239,17 +279,18 @@ def test_run_coverage_doctor(self):
             self.assertEqual(dca.run_coverage_doctor(dataset_id, data_product_id=data_product_id), 'Repair Successful')
 
         # Stream some data to the coverage
-        streamer = Streamer(data_product_id, interval=0.5, simple_time=True)
```
lukecampbell, line 242: So, my guess is that timeout had something to do with `simple_time`

blazetopher, line 242: possibly, but I got rid of that yesterday and it was still happening....it's also in another test and the Timeout doesn't happen

ion/util/direct_coverage_utils.py
blazetopher: replaced with sys.stderr.  not logging because the utility may be used from the command line (i.e. via manhole) where logging is not available.

PR #1190, opened 2013-06-11
Roll back Enhanced Resource Registry Client changes that break tests

ianfixes:
Code was added that prevents the cache from being updated.  This caused several instances of the cache becoming stale in INT test code.  This commit removes the errant code.

### DIFF
--- a/ion/util/enhanced_resource_registry_client.py
+++ b/ion/util/enhanced_resource_registry_client.py
@@ -330,6 +330,7 @@ class EnhancedResourceRegistryClient(object):
         object_id, object_type = self._extract_id_and_type(object)
 
         if not self.has_cached_predicate(predicate):
+            log.debug("Cache miss (predicate %s)", predicate)
             ret, _ = self.RR.find_subjects(subject_type=subject_type,
                                            predicate=predicate,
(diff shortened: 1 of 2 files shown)


ion/util/enhanced_resource_registry_client.py
```diff
@@ -496,57 +499,71 @@ def advance_lcs(self, resource_id, transition_event):
...
 
-    def filter_cached_associations(self, predicate, is_match_fn):
-        if not self.has_cached_predicate(predicate):
-            raise BadRequest("Attempted to filter cached associations of uncached predicate '%s'" % predicate)
+    def filter_prefetched_associations(self, predicate, is_match_fn):
+        if not self.has_prefetched_predicate(predicate):
+            raise BadRequest("Attempted to filter prefetched associations of unfetched predicate '%s'" % predicate)
 
-        return [a for a in self._cached_predicates[predicate] if is_match_fn(a)]
+        return [a for a in self._fetched_predicates[predicate] if is_match_fn(a)]
```
lukecampbell: You could get a speed boost here with

```
return filter(is_match_fn, self._fetched_predicates[predicate])
```

PR #1192, opened 2013-06-11
Fix Direct Access Timeout Issue

wfrench:
Direct access timeouts were not notifying the instrument agent that the DA server was brought down.  I've added tests to exercise all the DA terminal cases in the SBE37 driver:

https://github.com/wfrench/marine-integrations/blob/master/mi/instrument/seabird/sbe37smb/ooicore/test/test_driver.py#L1305


edwardhunter:
Looks good.

PR #1196, opened 2013-06-12
Uses REGEX to identify uuids instead of exceptions

lukecampbell:
Fixes [OOIION-1096](https://jira.oceanobservatories.org/tasks/browse/OOIION-1096) Pubsub checks for resource by catching exception, needs to use regex


blazetopher:
:+1:

mmeisinger:
:cookie:

PR #1203, opened 2013-06-14
let test retry if initial client cannot connect

newbrough:
intermittent error causes test to fail b/c agent does not start before service call returns.  this change allows the client connection to fail and will retry up to 5min.


ion/agents/data/handlers/test/test_instrument_end_to_end.py
```diff
@@ -104,7 +107,15 @@ def do_read_dataset(self):
         # should i wait for process (above) to start
         # before launching client (below)?
         #
-        self.client = ResourceAgentClient(self.device._id, process=FakeProcess())
+        self.client = None
+        end = time.time() + MAX_AGENT_START_TIME
+        while time.time()<end:
+            try:
+                self.client = ResourceAgentClient(self.device._id, process=FakeProcess())
```
carueda, line 114: are you probably missing a break right after this?

PR #1210, opened 2013-06-18
Decomposes the test_observatory test into multiple sets of assertions

lukecampbell:
This is a work in progress...


blazetopher:
:+1:

PR #1227, opened 2013-06-21
fix deployment extension for UI

newbrough:
not 100% complete -- still need to calculate status of attribute sites and devices,
but want to merge this much now b/c resource.py change made in last few days
changes OMS.get_deployment_extension to raise exception where before it just exited early b/c of a None object.


mmeisinger:
nice. looks good to me

PR #1233, opened 2013-06-25
Skips launching the qc post processor if test

lukecampbell:
This is a naive solution that skips the qc post processor
creation if the string 'test' is in the system name.


blazetopher:
:+1:

PR #1234, opened 2013-06-25
Improved direct access disconnect logic

wfrench:
Added logic to stop DA server as soon as we are notified.  Working hard to reduce the race condition possibilities. We will open another jira task as a minor to completely address the race condition.


mmeisinger:
Looks good to me

PR #1236, opened 2013-06-25
raise Unauthorized if user tries to view anothers UserInfo

newbrough:


pkediyal:
I was still reviewing it .. you merged it already !

PR #1237, opened 2013-06-25
Serial port_agent_process mode last mile of code.

unwin:
This is the last bit of code needed to get port_agent to start up configured to talk to a serial port instrument.  Have discussed/shared the change with Bill F. and David E.


mmeisinger:
:+1: thanks for the cleanup

PR #1238, opened 2013-06-25
extended deployment service impl

newbrough:
service implementation to populate deployment extension object
TODO: need status rollup of instruments, platforms, portals


mmeisinger:
thanks for the cleanup

PR #1246, opened 2013-06-28
OOIION-969: Process dispatcher doesn't handle EEAgent messaging errors

labisso:
Most of the fix is in the underlying epu repo


mmeisinger:
Thanks!

PR #1256, opened 2013-07-14
fix OOIION-984; allow instrument life-cycle changes by observator_operat...

pkediyal:
ran smoke and unit tests, should be good to go.

Note: ALL RESTRICTIONS RELATED TO LCS-CHANGE OPERATIONS HAVE BEEN LIFTED FOR OBS_OPERATOR and ORG_MANAGER


mmeisinger:
For the given set of desired LCS policy, the solution seems correct. Let's see how this works out...

PR #1264, opened 2013-07-19
OOIION-1152; only allow members of facility to subscribe to resources

pkediyal:


pkediyal:
fixes JIRA issue https://jira.oceanobservatories.org/tasks/browse/OOIION-1152

mmeisinger:
Looks good.
Please update this page and add the new policy to this table (and any ones that you added before without updating the page)

PR #1266, opened 2013-07-22
OOIION-1151; permit users to edit their account settings

pkediyal:
ran nosetests (unit and smoke), there wasn't a need to update policy documents


pkediyal:
Fixes issue: https://jira.oceanobservatories.org/tasks/browse/OOIION-1151

PR #1273, opened 2013-08-05
Fixed Length Attribute Strings and HDF5 Debugging

lukecampbell:


lukecampbell:
Tests pass: http://buildbot.oceanobservatories.org:8010/builders/coi_smoke/builds/2722

PR #1280, opened 2013-08-12
Renames module, adds more assertions and logging

lukecampbell:
This commit expands the compatibility of this utility a bit. For some reason the standard HDF library is allowed to have more messages listed in the Object Header message than actually appear in the object list. The way I got around this was to add a local context EOF to the context objects, which is updated by object headers and continuation messages. If the local offset > local context EOF then it's one of these cases where there are more messages indicated than actually listed. 

I also moved all the sys.stdout writes to logging statements so that the handler and output can be customized or ignored. 

The extra assertions now check for version compatibility and various values, if the values are blatantly wrong in the superblock this should catch it.

### DIFF
--- a/ion/util/h5debug.py
+++ b/ion/util/h5debug.py
@@ -4,11 +4,12 @@
 @file ion/util/h5_graph.py
 '''
 
-import sys
 from struct import unpack
-import numpy as np
 from StringIO import StringIO
+import os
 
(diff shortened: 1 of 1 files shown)


lukecampbell:
Purely Additive, the module is not incorporated anywhere, no buildbot test required.

PR #1285, opened 2013-08-13
Fixes double-launch from QC Processing Tests

lukecampbell:


lukecampbell:
Tests Pass [here](http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/86)

PR #1291, opened 2013-08-20
OOIION-1232; permit registered users to publish reporting_issue_event

pkediyal:
I have changed the solution from what you had suggested (i.e., create a wrapper function publish_issue_event for publish_event when event_type = ResourceIssueReportedEvent)

Instead, I make that check for (event_type = 'ResourceIssueReportedEvent' ) in the policy

Here is the JIRA issue: https://jira.oceanobservatories.org/tasks/browse/OOIION-1232


mmeisinger:
That looks good

PR #1295, opened 2013-08-23
Changes default h5py opening mode from r+ to r

lukecampbell:
The default opening mode is 'r+' which is actually read/write, this patch
changes the default behavior so that when files are opened with no mode
explicitly specified they are opened in read-only without the use of an
exclusion lock.

This patch may cause unforeseen issues if readers depend on write privileges;
there is no evidence to suggest that they do.


lukecampbell:
SMOKE, and a large test-suite pass. Coverage Model passes as well.

PR #1299, opened 2013-08-24
Updated discovery_service.py to add limits to the search results

seman:


lukecampbell:
The discovery service already allows for explicit limits, this code adds a conflicting API to the interactions of the discovery API. 

I just added an example that shows how to use the LIMIT keyword in the query string which will do what you want. A limit keyword component already exists for the query.

https://github.com/ooici/coi-services/blob/master/ion/services/dm/presentation/test/discovery_test.py#L601

PR #1314, opened 2013-09-09
OOION-1245; enhance create_attachment policy

pkediyal:
these are purely policy changes but require code to be written in resource_registry. please review.

current policy only allows attachment to orgs, UserInfo, and resources that have an "hasResource" association with a relevant org,


pkediyal:
hold on let me add DATA_OPERATOR role to the code; I had thought DATA_OPERATOR is not a role in R2 and so had not included it, but I just saw that on ion-alpha DATA_OPERATOR is a role that can be offered.

PR #1316, opened 2013-09-10
Fixes for OOIION-1276, dataset agent timestamp issues

sfoley:
This pull request improves testing of the particle-to-granule code in coi-services that makes the link between MI particles and the ION granules. It also corrects issues with the wrong timestamp being loaded for particles that use the port timestamp or the internal timestamp for the preferred timestamp. This will definitely affect dataset agent granules, possibly other data sources as well.

### DIFF
--- a/ion/agents/populate_rdt.py
+++ b/ion/agents/populate_rdt.py
@@ -12,10 +12,33 @@ __license__ = 'Apache 2.0'
 
 import numpy
 import base64
+from pyon.public import log
+from ion.agents.data.parsers.parser_utils import DataParticleKey
 
 def populate_rdt(rdt, vals):
-    
-    
+    """
+    Populate a RecordDictionaryTool object with values from a data particle
+    @param rdt An empty/fresh RecordDictionaryTool for the given stream
+    @param A list of data particle dictionaries to insert into the RDT.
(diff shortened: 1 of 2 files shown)


ion/agents/populate_rdt.py
```diff
@@ -36,7 +63,7 @@ def populate_rdt(rdt, vals):
                             value = base64.b64decode(value)
                         data_arrays[value_id][i] = value
             
-            elif k == 'driver_timestamp':
+            elif k == preferred_timestamp:
```
wfrench, line 66: What happens if the preferred timestamp has no value won't that break ingestion?  Reaching back in the memory banks I seem to recall talking about a progression for assigning the record timestamp.  I think it was something like, use the preferred timestamp if it existed otherwise user the driver timestamp because we are guaranteed that it will exists.  I don't know if that statement is still true since adding dataset agents, but it still seems we need some sort of progression to ensure each granule has a timestamp.

sfoley:
The particle structure itself always generates a timestamp. When it publishes, it verifies one exists. Even so, this code assumes that, with no preferred timestamp, the driver timestamp gets used (which should always be there, even if there is no internal time or a port agent involved).

PR #1318, opened 2013-09-10
enhance create_attachment policy

pkediyal:
these are purely policy changes but require code to be written in resource_registry. please review.

current policy only allows attachment to orgs, UserInfo, and resources that have an "hasResource" association with a relevant org,


mmeisinger:
looks good. merge when ready

PR #1341, opened 2013-09-23
Fixes two data product related tests

lukecampbell:
Fixes the following failed tests:
- `ion.services.dm.inventory.test.test_dataset_management:DatasetManagementIntTest.test_dataset_crud`
- `ion.services.sa.product.test.test_data_product_management_service_integration:TestDataProductManagementServiceIntegration.test_derived_data_product`

register_dataset no longer takes advantage of the registered attribute
for a dataset resource, so the test no longer needs to verify that.

activate_data_product on a derived data product is now automatic during
creation. Calling activate_data_product on a derived data product
results in a BadRequest being raised, so this test got patched.


lukecampbell:
Submitted for review

PR #1344, opened 2013-10-02
Ooiion1277 r2

wbollenbacher:
Fix for OOIION-1277 for R2
@edwardhunter Please review


jamie-cyber1:
Do not merge. Needs to be cleaned up.

PR #1367, opened 2013-10-09
Fixed publishing logic to keep key-value pair of preferred timestamp in ...

edwardhunter:
Fixes 1328:
https://jira.oceanobservatories.org/tasks/browse/OOIION-1328


edwardhunter:
@jamie-cyber1 please review!

PR #1372, opened 2013-10-14
OOIION-1327 Change assert to PlatformException to enable debug

MauriceManning:
@jamie-cyber1 please review

### DIFF
--- a/ion/agents/platform/platform_agent.py
+++ b/ion/agents/platform/platform_agent.py
@@ -2311,7 +2311,7 @@ class PlatformAgent(ResourceAgent):
                 children were processed OK.
         """
         instrument_ids = self._get_instrument_ids()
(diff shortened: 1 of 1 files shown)


ion/agents/platform/platform_agent.py
```diff
@@ -2311,7 +2311,7 @@ def _instruments_shutdown_and_terminate(self):
                 children were processed OK.
         """
         instrument_ids = self._get_instrument_ids()
-        assert instrument_ids == self._ia_clients.keys()
+        raise PlatformException("%s does not equal to %s ", repr(set(instrument_ids)), repr(set(self._ia_clients.keys() ) ) )
```
carueda: you are raising the exception unconditionally?  seems like you're missing a condition like:
   if set(instrument_ids) != set(self._ia_clients.keys()):
      raise ....

PR #1377, opened 2013-10-15
OOIION-1342. Remove two unused computed attributes

mmeisinger:
Bump ion-definitions pointer


mmeisinger:
@MauriceManning please review

PR #1393, opened 2013-10-23
OOIION-1356. Clear out agent_config in agent_instance

mmeisinger:
of device extended resource


mmeisinger:
@jamie-cyber1 please review

PR #1400, opened 2013-10-23
OOIION-1359 Dataset with 22k timesteps takes greater than 20s to retrieve

lukecampbell:
Fixes slowness in Coverage -> RDT
Values coming directly from the coverage model are already vetted and
are capable of being safely stored in another coverage so we don't ened
to rely on the RDT checks and verification subroutines, we can just set
teh value arrays directly. Major speed increase.

Part of [OOIION-1359](https://jira.oceanobservatories.org/tasks/browse/OOIION-1359)


lukecampbell:
DM Tests [here](http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/122)

PR #1403, opened 2013-10-24
OOIION-1373: Make policy (re)load order deterministic by applying create order /master

mmeisinger:


mmeisinger:
@jamie-cyber1 please review

PR #1415, opened 2013-10-28
Adds more context information to logging exceptions in viz

lukecampbell:
- Adds better context logging in data retriever
- Adds better context logging in visualization


lukecampbell:
@jamie-cyber1 Ready

PR #1420, opened 2013-10-29
OOIION-1346 Adds more detailed logging to QC Post Processor

lukecampbell:
Part of [OOIION-1346](https://jira.oceanobservatories.org/tasks/browse/OOIION-1346)

This commit doesn't patch any code, but hopefully it will allow me to
narrow down what is happening.


lukecampbell:
@jamie-cyber1 It's probably safe to merge any time, it only adds logging and exception handling if a BadRequest is caught from the data retriever client.

PR #1430, opened 2013-11-06
Adds test for CTDBP-NO

lukecampbell:
There's no JIRA bug for this, it's just a utility to assist in future debugging of preload issues.


lukecampbell:
@jamie-cyber1 Ready

PR #1434, opened 2013-11-06
May address XML comparison bug on CentOS machines

lukecampbell:
Part of [OOIION-1412](https://jira.oceanobservatories.org/tasks/browse/OOIION-1412)


lukecampbell:
@jamie-cyber1 Ready

PR #1437, opened 2013-11-07
OOIION-1378: OOI preload with full data product generation support /REL2.0

mmeisinger:
- Enhance OOI preload to be idempotent and to deal with data products idempotently
- Updates ion_loader to separate bulk insert from update
- Updated OOI preload to support platform data product generation
- Fix preload tests
- Small fix to ContainerUI menu


mmeisinger:
@jamie-cyber1 please review

PR #1440, opened 2013-11-08
OOIION-1358: Move agent config to object in object store /master

mmeisinger:
Need to merge pyon first


mmeisinger:
@jamie-cyber1 please review

PR #1443, opened 2013-11-08
OOIION-1358: Move agent config to object in object store /master

mmeisinger:


mmeisinger:
Hold this until test fix. The issue is that the container object_store returns a pretty bare dict from couchdb. It uses the couchdb json, which is standard json. It returns unicode strings in many cases

MauriceManning:
:thumbsup:

PR #1444, opened 2013-11-09
OOIION-1353: Provide prior agent state to new launch of dataset agent /master

mmeisinger:
merge ion-definitions first


mmeisinger:
@jamie-cyber1 please review, ion-definitions first

PR #1445, opened 2013-11-09
OOIION-1353: Provide prior agent state to new launch of dataset agent /REL2.0

mmeisinger:
merge ion-definitions first


mmeisinger:
@jamie-cyber1 please review, ion-definitions first

PR #1457, opened 2013-11-15
Fix to make preload for external dataset agent idempotent /REL2.0

mmeisinger:


mmeisinger:
@jamie-cyber1 please review

PR #1490, opened 2013-11-21
OOIION-1358: Switch all agent launcher to use object store for config

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #1494, opened 2013-11-21
Fixes key error for dataset_temporal_bounds

lukecampbell:
The metadata queries now return an empty dict if there is no data.

Associated with the email thread: `Getting Error when trying to retrieve an empty coverage?`
No Jira bug assigned.


lukecampbell:
@jamie-cyber1 Ready to merge after tests.

PR #1512, opened 2013-11-22
Remove sha checking until it is fixed

jamie-cyber1:


jamie-cyber1:
@mmeisinger please review.

PR #1516, opened 2013-11-25
OOIION-1500 Updates URL for griddap->tabledap

lukecampbell:
https://jira.oceanobservatories.org/tasks/browse/OOIION-1500


lukecampbell:
@jamie-cyber1 At your leisure, we'd like to get these into alpha as soon as convenient.

PR #1534, opened 2013-11-26
IMS reads object store for instrument device status when availabe. Makes...

edwardhunter:
... face pages faster. OOIION-1396


mmeisinger:
Looks good! A few minor comments:
- There is at least one print statement remaining
- It may be more robust to use state_dict.get("agg_status", {}) instead of relying on the fact it is there
- dict.has_key() is deprecated. Use if device_id in dict instead

PR #1560, opened 2013-12-04
ooiion-1374 Platform agent must manage port power for child devices : add test and defaults 

MauriceManning:
Updates for this bug after additional testing 
- add test (must run locally) for RSN hardware
- set default for driver_config.attributes 
- add a recursion param to go_inactive

@jamie-cyber1 please review
@crueda please review


carueda:
Looks OK to me.  Since the key test test_platform_with_instrument_streaming is skipped, I think please go ahead and merge; I can then run the test locally.

PR #1597, opened 2013-12-09
OOIION-1550: Add devices to find_relates_sites

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge ion-defs then I update this

PR #1616, opened 2013-12-12
fix OOIION-1587

carueda:
https://jira.oceanobservatories.org/tasks/browse/OOIION-1587

systematic  "except\s*:" --> "except Exception:" replacement.


carueda:
@jamie-cyber1 please review
@edwardhunter please review

carueda:
The replacement was systematic but manually applied for further inspection in the code. No impact whatsoever on the functional behavior of the tests, the reason simply being that the except statement is now more specific than before, so if no issues were encountered with the empty statement, then much less with the more specific one.

This is a general improvement in the code.  No specific broken tests were addressed with this change. 

All platform unit and integration tests completing ok.
No specific script for UI testing.

PR #1618, opened 2013-12-12
fix OOIION-1318

carueda:
https://jira.oceanobservatories.org/tasks/browse/OOIION-1318


carueda:
@edwardhunter please review
@jamie-cyber1 please review

PR #1626, opened 2013-12-18
fix OOIION-1608 list index out of range

carueda:
https://jira.oceanobservatories.org/tasks/browse/OOIION-1608

The test test_13_platforms_and_01_instruments now completes fine. But note that this test (and others in the same file) are currently skipped by default per OOIION-1297.


carueda:
@edwardhunter and/or @jamie-cyber1 please review

PR #1645, opened 2014-01-06
OOIION-1623 Remove CheckSync from platform driver interface 

MauriceManning:
Remove CheckSync from platform driver interface until this capability is fully defined with RSN/OMS team.

@edwardhunter please review

@carueda please review


edwardhunter:
Maurice this has 250+ commits...I don't think it can be merged. Can you resubmit with one commit.

PR #1648, opened 2014-01-06
OOIION-1655 Remove current impl of register_instrument_agent.

MauriceManning:
Will be implemented in R3 as as part of D041 Data Processing Management Service

@jamie-cyber1 please review


jamie-cyber1:
How is NotImplementedError imported? This should be accompanied by a UNIT test that calls this function and verify the error was thrown.

jamie-cyber1:
:+1:

PR #1652, opened 2014-01-07
OOIION-1617: Use alternate instrument class name for preload generated model, site, device

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #1659, opened 2014-01-08
OOIION-1352 add oms url to preload

MauriceManning:
add oms url as a column in the preload spreadsheet, fill only when it is not using the sim.

@mmeisinger please review


mmeisinger:
This will break OOI preload. Please add the row "oms_url" to the _load_PlatformAgentInstance function or use row.get("oms_url", "") instread

mmeisinger:
:+1: tests exist

PR #1660, opened 2014-01-08
OOIION-1352 add oms url to preload

MauriceManning:
add oms url as a column in the preload spreadsheet, fill only when it is not using the sim.

@mmeisinger please review


mmeisinger:
:+1:

PR #1666, opened 2014-01-09
ooiion-1646 Add data product persist events

MauriceManning:
Add events when data product persist is turned on or off. Also fix an error when suspend persistence was called on an unpersisted data product, log a warning in this case.

@mmeisinger please review


mmeisinger:
:+1:

PR #1675, opened 2014-01-15
OOIION-1670 Command Line Utility for Applying Calibration Coefficients R2

lukecampbell:
There's a test included in this pull request that acts as a regression test for this capability but the true measure of verification will be if an operator can inject value sets into a running system.


lukecampbell:
@MauriceManning ready to merge.

PR #1688, opened 2014-01-23
Bump pyon pointer to include test gevent block changes

jamie-cyber1:


jamie-cyber1:
@wbollenbacher please merge after tests pass.

PR #1689, opened 2014-01-23
Bump pyon to include gevent block test changes

jamie-cyber1:


jamie-cyber1:
@wbollenbacher please merge after tests pass.

PR #1691, opened 2014-01-24
TEST - Adds convenience function to test harness for DM R2

lukecampbell:
This adds a convenience function that is used by the integration tests for some of the new agents. This does not affect any current functionality and has no impact on the system, it only eases future integration testing.


lukecampbell:
@jamie-cyber1 Can you review and merge please?

PR #1699, opened 2014-01-28
Adjust test error count to account for skipped error condition.

edwardhunter:
Bill French skipped an error test but did not adjust the assert count of error events. We need to make sure this error that is currently skipped behaves properly.


edwardhunter:
Appears the MI Logic has been fixed so I will cancel this request and send a new one with the correct, uncommented test logic.

PR #1701, opened 2014-01-28
Fixes command_errror_test in instrument agent.

edwardhunter:
Uncomment command error block that now returns conflict from MI process correctly.


edwardhunter:
@jamie-cyber1 review.

PR #1703, opened 2014-01-28
CIDEVSTC-49 Adds prototype packager for processing eggs

lukecampbell:
M087 Data Processing workflow management

This is the first iteration of a packager utility that can package process-based eggs.


lukecampbell:
@MauriceManning Can you review this too? It's probably not going to make it but at least if the code is merged it can be used as a reference.

PR #1704, opened 2014-01-28
OOIION-1697: Discovery should not return retired resources

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge pyon first

PR #1712, opened 2014-01-30
OOIION-1719 OMS Service caches associations between service calls

jamie-cyber1:
Remove ObservatoryUtil service attribute from OMS service to avoid incorrect caching of resources and associations between service calls


jamie-cyber1:
@mmeisinger please review merge.
Note please look at the first commit for substantive changes.  The 2nd commit removes DOS formatting from the entire file.
FYI, @MauriceManning and @edwardhunter on the issue and the fix.

mmeisinger:
:+1: Good catch and rework

PR #1718, opened 2014-01-30
Added event when DSA opens a new file OOIION-1672

wfrench:
@edwardhunter can you please review this?

Added interface so the driver can raise an event when a new file is opened.  This will allow us to determine the source file for samples read.


edwardhunter:
Bill, got a failed test build here.
Plus please add comment on how this was tested.

wfrench:
What test failed?  

Tested in coi with:
https://github.com/ooici/coi-services/blob/master/ion/agents/data/test/test_dsa_moas_ctdgv.py

Tested in mi with:
https://github.com/ooici/marine-integrations/blob/master/mi/dataset/driver/moas/gl/ctdgv/test/test_driver.py

PR #1719, opened 2014-01-30
Added event when DSA opens a new file OOIION-1672

wfrench:
@edwardhunter can you please review this?

Added interface so the driver can raise an event when a new file is opened. This will allow us to determine the source file for samples read.


edwardhunter:
Please add comment on how this was tested.

PR #1722, opened 2014-01-31
OOIION-1732 Adds ability for calibration injection to include preload identifiers

lukecampbell:


lukecampbell:
@jamie-cyber1 Ready to merge

PR #1726, opened 2014-01-31
Fix Dataset Agent Lost Connection Logic OOIION-1734

wfrench:
Dataset Agent no longer returning conflict on lost connection.  Added test to verify behavior.

https://github.com/wfrench/coi-services/blob/daaa1c0910c634e928b023f6b2697e8edcfe4a71/ion/agents/data/test/test_dsa_moas_ctdgv.py#L80


wfrench:
@edwardhunter can you check this out?  Thanks

PR #1738, opened 2014-02-03
CIDEVSTC-275 Creates a DataProcessDefinition for each ParameterFunction defined.

lukecampbell:


MauriceManning:
We usually dont have DM layer services create SA level resources. Can the DPD and DP be created by preload processing when the ParamFunction is created and the links made at that time?  If this is a short term workaround then thats ok as long as we clean up later. I should have the create_data_process_new and create_data_process_definition_new submitted in a PR in a day or two and they would also support creating a DPD with an assoc PF and DP.

lukecampbell:
OBE

PR #1739, opened 2014-02-04
CIDEVSTC-69	Integrate and test with production environment

edwardhunter:
Tests setup for cabled infrastructure support. Still required actual test logic to simulate UI workflow.


edwardhunter:
@jamie-cyber1 
Please review.

PR #1740, opened 2014-02-05
CIDEVSTC-43	 Implement data process management service extensions

MauriceManning:
add transform worker and test as well as new create methods for data process and data process definition

@lukecampbell 

Initial impl of transform worker as assoc test. Initial impl of new methods to create data process and data process definition in DPMS.


lukecampbell:
Merged into my branch and PR submitted.

PR #1742, opened 2014-02-05
CIDEVSTC-38 - Adds egg-download for TransformWorker and Test

lukecampbell:
- Downloads egg into a temporary directory (consistent throughout the
  container)
- If the URL can't be loaded, it fails.

### DIFF
--- a/ion/processes/data/transforms/transform_worker.py
+++ b/ion/processes/data/transforms/transform_worker.py
@@ -23,6 +23,9 @@ from pyon.core.object import IonObjectSerializer
 from pyon.ion.stream import StreamSubscriber
 from pyon.ion.event import handle_stream_exception
 
+from tempfile import gettempdir
+import os
+import requests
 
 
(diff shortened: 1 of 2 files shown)


ion/processes/data/transforms/transform_worker.py
```diff
@@ -196,15 +197,25 @@ def create_publishers(self):
...
-        return filename
+        if r.status_code == 200:
+            # Download the file using requests stream
+            with open(path, 'wb') as f:
+                for chunk in r.iter_content(chunk_size=1024):
+                    if chunk:
+                        f.write(chunk)
+                        f.flush()
+            return path
+        raise IOError("Couldn't download the file at %s" % url)
```
lukecampbell, line 220: I thought about putting the eggs into the shared filesystem but I realized that if multiple containers try to contend for loading eggs at the same time from a shared filesystem while one may download it it could cause bad behavior. I opted for the simplest solution: do what MI does.

PR #1743, opened 2014-02-06
CIDEVSTC-43	 Implement data process management service extensions

MauriceManning:
add status event check, refine test, fix DPD create issue

@lukecampbell please review

### DIFF
--- a/ion/processes/data/transforms/transform_worker.py
+++ b/ion/processes/data/transforms/transform_worker.py
@@ -5,12 +5,13 @@
 @description Data Process worker
 '''
 from interface.services.coi.iresource_registry_service import ResourceRegistryServiceClient
-from interface.objects import Granule
+from interface.objects import Granule, DataProcessStatusType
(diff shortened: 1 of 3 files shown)


ion/processes/data/transforms/transform_worker.py
```diff
@@ -54,6 +58,7 @@ def on_start(self): #pragma no cover
 
         TransformStreamProcess.on_start(self)
 
+        #todo: can the subscription be changed or updated when new dataprocesses are added ?
```
lukecampbell, line 61: A specific subscription can be suspended or deleted.
New subscriptions can be added no problem but for every new subscription we need to have a new stream subscriber object. A pool of them. Maintaining the consistency between the subscription resources and the subscribers should be coordinated through data process management.

i.e.

```
add_subscription(data_process, stream)
  for each data process in some pool:
    tell process to add subscriber
    create subscritption and activate
```

lukecampbell:
@MauriceManning I'll merge when the test is complete.

PR #1745, opened 2014-02-06
TEST - Fixes test_transform_worker

lukecampbell:
Fixes the broken test.
- Places EGG URI in the info record
- Downloads egg at import phase

The test failed because of `ImportError: No module named ion_example.add_arrays`
This PR fixes that issue.


lukecampbell:
Here's the test on coi_any

http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/156

PR #1752, opened 2014-02-07
Fix gov test

jamie-cyber1:


jamie-cyber1:
@mmeisinger please review.
Michael, I've modified the test to match the code for the current instrument extended attribute state.
Please check to make sure my assumptions are correct.

Adding @pkediyal and @edwardhunter so they are aware.

PR #1757, opened 2014-02-07
OOIION-1741: Remove ElasticSearch support from system

mmeisinger:
- Polish discovery
- Add discovery service tests


mmeisinger:
ion-definitions pointer has to be updated before merge

ion/processes/bootstrap/plugins/bootstrap_index.py
```diff
@@ -13,15 +13,5 @@ class BootstrapIndex(BootstrapPlugin):
...
-            config = DotDict(config)
-            config.op                   = 'clean_bootstrap'
-
-            process.container.spawn_process('index_bootstrap','ion.processes.bootstrap.index_bootstrap','IndexBootStrap',config)
-            #---------------------------------------------
-        else:
-            log.info("Not creating the ES indexes.")
-
+        # There used to be initialization of ElasticSearch, now removed by using Postgres
+        pass
```
lukecampbell, line 17: I think we could probably get rid of this bootstrap_index.py

ion/services/dm/presentation/test/discovery_test.py
```diff
@@ -445,213 +418,23 @@ def setUp(self):
...
-    def test_associative_searching(self):
+        pd_id, _ = self.rr.create(pd)
 
-        dp_id,_ = self.rr.create(DataProduct('test_foo'))
-        ds_id,_ = self.rr.create(Dataset('test_bar', registered=True))
-        self.rr.create_association(subject=dp_id, object=ds_id, predicate='hasDataset')
+        search_string = "search 'index_location' geo distance 20 km from lat 0 lon 0 from 'devices_index'"
 
-        search_string = "search 'type_' is 'Dataset' from 'resources_index' and belongs to '%s'" % dp_id
+        results = self.poll(9, self.discovery.parse,search_string)
```
lukecampbell, line 430: We had to use a polling mechanism with ES because it wasn't immediately consistent with CouchDB. With PostgreSQL consistency is guaranteed and immediate so polling isn't necessary, it can be provably argued that if the resource can be found after it's created it will be found the first time.

lukecampbell:
If you rebase these on the latest master we can rerun the tests and merge.

mmeisinger:
I have rebased and force pushed with the updated master pointer

lukecampbell:
Excellent, will merge shortly.

ion/processes/bootstrap/plugins/bootstrap_index.py
mmeisinger, line 17: I left it in in order to not throw over deploy and launch scripts that happen to reference this one. I agree it should probably go away and I did not find any more references to this one

lukecampbell:
One thing to note, and it's not directly part of M112 or any known action items, is that UNS has no integration tests that run, which probably means it doesn't work.

PR #1767, opened 2014-02-11
CIDEVSTC-278 Adds inspection for data process definitions

lukecampbell:
- Adds method in data_process_management for inspecting the source code of a data process definition
- Adds test for testing the inspection of a well-known dpd and source code
- Refactors TestDMExtended's SA-based tests into a new test class TestDataProcessFunctions

Depends on https://github.com/ooici/ion-definitions/pull/486

### DIFF
--- a/ion/services/sa/process/data_process_management_service.py
+++ b/ion/services/sa/process/data_process_management_service.py
@@ -111,11 +111,12 @@ class DataProcessManagementService(BaseDataProcessManagementService):
                tf1.uri           == tf2.uri    and \
                tf1.function_type == tf2.function_type
 
-    def create_transform_function(self, transform_function=''):
+    def create_transform_function(self, transform_function=None):
         '''
         Creates a new transform function
         '''
-
+        if not transform_function:
+            raise BadRequest("Improper TransformFunction specified")
         return self.RR2.create(transform_function, RT.TransformFunction)
 
     def read_transform_function(self, transform_function_id=''):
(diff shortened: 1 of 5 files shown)


lukecampbell:
Ready to merge.
here's the test on coi_any
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/168

ion/services/sa/process/data_process_management_service.py
```diff
@@ -392,6 +394,45 @@ def create_data_process_new(self, data_process_definition_id='', in_data_product
...
+        Returns the source code for the data process definition 
+        '''
+        dpd = self.read_data_process_definition(data_process_definition_id)
+        if dpd.data_process_type == DataProcessTypeEnum.PARAMETER_FUNCTION:
+            return self._inspect_parameter_function(data_process_definition_id)
+        #TODO: add the rest of the types
+        else:
+            raise BadRequest("Data process type not yet supported")
+
+    def _inspect_parameter_function(self, data_process_definition_id):
```
MauriceManning, line 411: this could just be inspect_function which checks the type of the DPD then follows the correct assoc to get either PF or TF. The logic should be the same for both.

lukecampbell, line 411: There's a bit of a difference. The TF has all of the module and egg information defined in the resource. The PF has all the information defined in a serialized dictionary for the Coverage Model.

MauriceManning, line 411: yes, agreed.

ion/services/sa/product/data_product_management_service.py
```diff
@@ -149,7 +149,9 @@ def create_data_processes(self, data_product_id=''):
             dp = DataProcess()
             dp.name = 'Data Process %s for Data Product %s' % ( dpd.name, data_product.name )
             # TODO: This is a stub until DPD is ready
-            self.clients.resource_registry.create(dp)
+            dp_id, _ = self.clients.resource_registry.create(dp)
+            self.clients.resource_registry.create_association(dpd._id, PRED.hasDataProcess, dp_id)
+            self.clients.resource_registry.create_association(dp_id, PRED.hasOutputProduct, data_product._id)
```
lukecampbell, line 154: This is important.

For every parameter function in every data product:
- A data process is created. 
- The DPD is associated with the DP
- The DP is associated with the Data Product.

PR #1769, opened 2014-02-11
Preventing a test from hanging the entire suite indefinitely until driver is fixed

jamie-cyber1:


jamie-cyber1:
@edwardhunter please review and merge.

PR #1773, opened 2014-02-12
CIDEVSTC-99: Directory service enhancements for count and skip

brianmckenna:
M112 - T131 SKIP parameter and total results in DB

M112-T131-A002 Return total results available in DB
M112-T131-A005 Process offset (skip) parameter for pagination
M112-T134-A002 Testing skip/offset parameter, implemented with M112-T131-A005
M112-T134-A002 Testing return count, implemented with M112-T131-A002

### DIFF
--- a/ion/services/dm/presentation/discovery_service.py
+++ b/ion/services/dm/presentation/discovery_service.py
@@ -35,31 +35,33 @@ class DiscoveryService(BaseDiscoveryService):
     # Query Methods
     #===================================================================
 
-    def query(self, query=None, id_only=True, count=False):
+    def query(self, query=None, id_only=True, search_args={}):
         """Issue a query against the indexes as specified in the query, applying filters and operators
         accordingly. The query format is a structured dict.
         See the query format definition: https://confluence.oceanobservatories.org/display/CIDev/Discovery+Service+Query+Format
 
         @param query    dict
         @param id_only    bool
+        @param search_args dict
         @retval results    list
         """
         validate_true(query, 'Invalid query')
 
-        return self.request(query, id_only, count=count)
+        return self.request(query, id_only, search_args=search_args)
 
-    def parse(self, search_request='', id_only=True, count=False):
+    def parse(self, search_request='', id_only=True, search_args={}):
         """Parses a given string request and assembles the query, processes the query and returns the results of the query.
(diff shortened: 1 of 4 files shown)


brianmckenna:
@mauricemanning @mmeisinger  ready for review

mmeisinger:
The operations parse and query are part of the YML service interface. You cannot add parameters here.

Please do not change the operation signature. Count is a special case. I would leave it away and do an id_only=True query instead.

mmeisinger:
To be more precise: it's good that you use id_only=True for the datastore query, but you need to indicate the count argument in the existing query arguments.

This means either extending the query DSL or adding a count: True key to the query intermediate format dict.

The rest of the changes looks good.

lukecampbell:
I think that we need to change the method signature to include this. I just evaluated what it would take to put it inside the query dictionary it's far simpler to accept it as an argument in the method.

Example

Our goal is to provide the UI with the count information: so the way I see this happening is that the UI makes two service calls. The first asking for the count of the query it's about to perform. The second is the query (with pagination). 

In the case of simple search (text box) the UI doesn't send a dictionary, which we would require if we didn't extend the method signature. Which has a consequence that we would need to extend the query language to add a new noun and update the parsing etc.  There's nothing wrong with this but I think it's just simpler to update the service method signature in lieu of updating the UI to append "COUNT" to every search string to get the count and updating the query language.

mmeisinger:
please add a search_args dict to the signature, not the count. count is a well known key in this search args dict, but this leaves room for other expansion.

We cannot afford to change the signature every so often

brianmckenna:
working on updating tests to reflect new return, once tests are modified will comment to request review/merge

lukecampbell:
Hold off on merging this PR until we merge https://github.com/ooici/ion-definitions/pull/487 and bump ion-defs.

brianmckenna:
ready for review and merge again. output kept as array, so no UI modifications necessary

mmeisinger:
Please update ion-definitions pointer. The changes are ready to merge afterwards

mmeisinger:
This PR did not contain the latest pointer to ion-definitions. The merge is now broken

lukecampbell:
I fixed it. Typo.

PR #1778, opened 2014-02-13

MauriceManning:
skip data process tests that are outdated in R3 and not used in R2

@jamie-cyber1  please review


lukecampbell:
she already skips the test with this PR: #1775
there's one for R2 as well.

PR #1785, opened 2014-02-14
Fix haagent test leaking policy thread loop

jamie-cyber1:
Fix and investigate why the policy_thread loop is not properly terminating sometimes.


jamie-cyber1:
@mmeisinger please review and merge.

PR #1793, opened 2014-02-18
Safer EEAgent heartbeat greenlet handling, updates OOIION-1753

daf:
The EEAgent heartbeat was using messaging and was being `.kill()`'d at any random point in execution on termination.  This results in possible full messaging/container failure.


daf:
@jamie-cyber1 @mmeisinger Please review.

mmeisinger:
Looks good to me. Is it confirmed that this should go into REL2.0 ?

daf:
Somone's going to have to tell me that.  It's def a problem that's been in there forever, so it affects 2.0.

PR #1800, opened 2014-02-19
OOIION-1756 Terminate high availability agent thread safely

jamie-cyber1:


jamie-cyber1:
@daf please review and merge.

daf:
Seems fine.  Could probably somehow combine these two event loops but this is a definite better approach than what we had.  Will merge when build passes.

jamie-cyber1:
Okay I will look into combining it.

daf:
Looking at it quickly I'm not sure there's a simple approach as they really are serving different purposes.  Maybe hold off for another PR?

jamie-cyber1:
Okay. Then go ahead and merge then.

PR #1801, opened 2014-02-19
OOIION-1756 Terminate high availability agent thread safely

jamie-cyber1:


jamie-cyber1:
@daf please review and merge.

daf:
Seems fine. Could probably somehow combine these two event loops but this is a definite better approach than what we had. 

Waiting for #1800.

jamie-cyber1:
Okay I will look into combining it.

PR #1809, opened 2014-02-19
OOIION-1758 Fix Timout import in port agent

jamie-cyber1:


jamie-cyber1:
@wfrench please merge after builder passes.

PR #1811, opened 2014-02-19
OOIION-1313 zmq send in driver client set to nonblocking

jamie-cyber1:


jamie-cyber1:
@edwardhunter please review and merge after build passes.

PR #1817, opened 2014-02-20
OOIION-1758 Check for non-existent pid file in port agent stop

jamie-cyber1:


jamie-cyber1:
@wfrench please review and merge.

PR #1828, opened 2014-02-21
Update pyon pointer to include OOIION-692

jamie-cyber1:


jamie-cyber1:
@edwardhunter please review and merge when builder passes.

PR #1830, opened 2014-02-21

MauriceManning:
add data process input/output map validation and test
add hasDataProductParent link between ins and outs of a TransformFunction

@jamie-cyber1 please review

@lukecampbell please review

### DIFF
--- a/ion/processes/data/transforms/test/test_transform_worker.py
+++ b/ion/processes/data/transforms/test/test_transform_worker.py
@@ -16,6 +16,7 @@ from pyon.event.event import EventSubscriber, EventPublisher
 from pyon.public import OT, RT, PRED
 from pyon.util.containers import DotDict
 from pyon.core.object import IonObjectDeserializer
+from pyon.core.exception import BadRequest
 from pyon.core.bootstrap import get_obj_registry
 from pyon.public import log, IonObject
 from interface.objects import  DataProcessTypeEnum, DataProcessTypeEnum, TransformFunctionType
@@ -98,6 +99,13 @@ class TestTransformWorker(IonIntegrationTestCase):
     @attr('LOCOINT')
     @unittest.skipIf(os.getenv('CEI_LAUNCH_TEST', False), 'Skip test while in CEI LAUNCH mode')
     def test_transform_worker(self):
+
+        # test that a data process (type: data-product-in / data-product-out) can be defined and launched.
+        # verify that the output granule fields are correctly populated
+
+        # test that the input and output data products are linked to facilitate provenance
+
+
         self.data_process_objs = []
         self._output_stream_ids = []
         self.event_verified = Event()
@@ -108,7 +116,7 @@ class TestTransformWorker(IonIntegrationTestCase):
         self.stream_def_id = self.pubsub_client.create_stream_definition(name='stream_def', parameter_dictionary_id=self.parameter_dict_id)
         self.addCleanup(self.pubsub_client.delete_stream_definition, self.stream_def_id)
 
-        # create the DataProduct
+        # create the DataProduct that is the input to the data processes
         input_dp_obj = IonObject(  RT.DataProduct, name='input_data_product', description='input test stream',
                                              temporal_domain = self.time_dom.dump(),  spatial_domain = self.spatial_dom.dump())
         self.input_dp_id = self.dataproductclient.create_data_product(data_product=input_dp_obj,  stream_definition_id=self.stream_def_id)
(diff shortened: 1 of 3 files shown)


ion/processes/data/transforms/test/test_transform_worker.py
```diff
@@ -145,9 +157,13 @@ def test_transform_worker(self):
 
         self.publisher.publish(rdt.to_granule())
 
-
+        #validate that the output granule is received and the updated value is correct
         self.assertTrue(self.event_verified.wait(10))
 
+        #validate that the code from the transform funcation can be retrieve via inspect_data_process_definition
+        src = self.dataprocessclient.inspect_data_process_definition(dataprocessdef_id)
+        self.assertTrue( 'def add_arrays(a, b)' in src)
```
lukecampbell: For future reference, the built-in unittest library comes packed with assertion methods to make your life easier. Also, when they fail they provide better messages for what part of the test failed.

http://docs.python.org/2/library/unittest.html#unittest.TestCase.assertEqual

for instance this line could be

``` python
self.assertIn('def add_arrays(a,b)', src)
```

ion/processes/data/transforms/test/test_transform_worker_subscriptions.py
```diff
@@ -162,6 +164,78 @@ def test_transform_worker(self):
...
+        rdt['time']         = [0] # time should always come first
+        rdt['conductivity'] = [1]
+        rdt['pressure']     = [2]
+        rdt['salinity']     = [8]
+
+        self.publisher_one.publish(msg=rdt.to_granule(), stream_id=self.stream_one_id)
+
+
+        self.assertTrue(self.event2_verified.wait(10))
+        self.assertTrue(self.event1_verified.wait(10))
```
lukecampbell: When using gevent-based Events to synchronize threads in a test and use for assertions, please make the wait time an environment/configurable variable so Jamie can increase the time.  Otherwise we'll have to keep making code changes to constantly increase the event timeouts as the system gets slower.

``` python
wait_time = CFG.get_safe('endpoint.receive.timeout', 10)
self.assertTrue(self.event1_verified.wait(wait_time))
self.assertTrue(self.event2_verified.wait(wait_time))
```

ion/services/sa/process/data_process_management_service.py
```diff
@@ -769,6 +750,62 @@ def _inspect_parameter_function(self, data_process_definition_id):
...
+
+        #load the associated transform function
+        if data_process_definition_obj.uri:
+            egg = self._download_egg(data_process_definition_obj.uri)
+            import pkg_resources
+            pkg_resources.working_set.add_entry(egg)
+        else:
+            log.warning('No uri provided for module in data process definition.')
+
+        module = importlib.import_module(tfunc_obj.module)
```
lukecampbell: there should probably be a try-block in here to catch an import error and then try to load the egg

``` python
try:
    module = importlib.import_module(tfunc_obj.module)
except ImportError:
    if not data_process_definition_obj.uri:
        raise
    egg = TransformWorker.download_egg(data_process_definition_obj.uri)
    import pkg_resources
    pkg_resources.working_set.add_entry(egg)
    module = importlib.import_module(tfunc_obj.module)

...
```

lukecampbell: Otherwise you're going to download and load an egg every time we call this method.

ion/services/sa/process/data_process_management_service.py
```diff
@@ -769,6 +750,62 @@ def _inspect_parameter_function(self, data_process_definition_id):
...
+
+        function = getattr(module, tfunc_obj.function)
+        # Gets a list of lines of the source code
+        src = inspect.getsourcelines(function)[0]
+        # Join the lines into one contiguous string
+        src = ''.join(src)
+
+        return src
+
+    def _download_egg(cls, url):
```
lukecampbell: Please remove this method. 

We can call the download method from [TransformWorker](https://github.com/lukecampbell/coi-services/blob/viz/ion/processes/data/transforms/transform_worker.py#L283-284)

That way we can keep the way we download the eggs consolidated; that way if we change it we won't have to update it in many places. Especially since the methods are identical.

ion/processes/data/transforms/test/test_transform_worker_subscriptions.py
MauriceManning: ok, then did you mean for the asserts to use 'wait_time' ?  as in:
self.assertTrue(self.event1_verified.wait(wait_time))

lukecampbell:
:+1:

PR #1835, opened 2014-02-24

MauriceManning:
- convert to new data process and data process definition create methods
- remove workflow from AS
- remove workflow from ion_loader and from sheets and workflow DPDs from data process definition sheet

@lukecampbell please review

@mmeisinger please review


mmeisinger:
:+1: Looks like a good cleanup to me. No issues.

PR #1836, opened 2014-02-25

MauriceManning:
remove obso code from DPMS
add data process events
move uri attribute from DPD to TF resource

@mmeisinger please review

@lukecampbell please review


mmeisinger:
:+1: looks good to me (on brief glance). Update ion-defs pointer and gio ahead when ready

PR #1837, opened 2014-02-25
STC M100 Marine Device Activation

edwardhunter:
Round trip test cases for rsn activation cycle.
@mmeisinger please review


edwardhunter:
@mmeisinger please review

mmeisinger:
:8ball: Excellent test and documentation. 

Here a few comments:
- Assigning a model to the device should be before the device is declared LCS.DEVELOPED.
- I wonder if there are other SA test that now can be removed because this encompassing test covers them

PR #1839, opened 2014-02-25
OOIION-1760 Fixes service-client None issue with discovery

lukecampbell:
Service clients don't pass empty dictionaries. They pass None (or null in JSON). Service implementors need to account for this by specifying keyword arguments that would normally default to an empty dictionary as None and then accepting either a dictionary object or None.

### DIFF
--- a/ion/services/dm/presentation/discovery_service.py
+++ b/ion/services/dm/presentation/discovery_service.py
@@ -69,7 +69,8 @@ class DiscoveryService(BaseDiscoveryService):
         query_request = parser.parse(query_string)
         return query_request
 
-    def request(self, query=None, id_only=True, search_args={}):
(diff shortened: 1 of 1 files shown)


lukecampbell:
I originally had this patch in another commit but I'll put it here so it gets merged under this bug and addressed sooner.

mmeisinger:
:+1: good

jamie-cyber1:
@lukecampbell you are submitting one for REL2.0 as well?

lukecampbell:
I don't know. Is this an issue on REL2.0?

lukecampbell:
I just looked at the REL2.0 code, it doesn't use this stuff and the interface is different. So, I don't think it's an issue for REL2.0.

jamie-cyber1:
Great. thanks.

PR #1840, opened 2014-02-25
CIDEVSTC-43 - Updates temporal bounds info in data products through ingestion

lukecampbell:


lukecampbell:
@MauriceManning at your leisure.

lukecampbell:
Here's the build
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/184

mmeisinger:
This is a pragmatic solution (and well implemented). Just an architectural note that this changes the nominal_datetime meaning to actual datetime. If something is automatically updated, it's not nominal anymore. Not sure this matters and if nominal is required by users, but it's an item to note. We are leaving the consistency of the OOI metadata model

lukecampbell:
What should we rename it to?

mmeisinger:
let's not rename it and just live with the inconsistency. a rename would cause ripples to the UI screens and UI database etc.
I made the comment for the record not as an action to change anything in this PR

lukecampbell:
ok.

@MauriceManning any concerns?

ion/processes/data/ingestion/science_granule_ingestion_worker.py
```diff
@@ -228,6 +237,43 @@ def update_metadata(self, dataset_id, rdt):
...
+
+    def update_geo(self, data_product, rdt):
+        lat = None
+        lon = None
+        for p in rdt.fields:
+            # TODO: Not an all encompassing list of acceptable names for lat and lon
+            if p.lower() in ('lat', 'latitude', 'y_axis'):
+                lat = np.asscalar(rdt[p][-1])
+            elif p.lower() in ('lon', 'longitude', 'x_axis'):
+                lon = np.asscalar(rdt[p][-1])
```
lukecampbell, line 266: I just want to point out to folks in this thread that we don't have a convenient way to determine if a variable refers to latitude or longitude. 

CF Standard relies on checking the name, standard name and units for well-known values. I don't have time to implement it fully that way but I think this will cover most cases.

PR #1844, opened 2014-02-25
OOIION-1768 Fix wrong event type published by dpms

jamie-cyber1:


jamie-cyber1:
@MauriceManning please review and merge.

PR #1846, opened 2014-02-26
TEST - Fixes edge case in ingestion where lat/lon are None

lukecampbell:
Fixes broken SMOKE build.


lukecampbell:
@jamie-cyber1 Ready for merge
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/189

PR #1848, opened 2014-02-26
OOIION-1774 Fixes ingestion bug and performance issue

lukecampbell:
Ingestion used to check to see if a parameter was empty by evaluating
it, which causes the entire parameter function to evaluate against the
array. This patch prevents ingestion from evaluating parameter functions
it doesn't need to.

[OOIION-1774](https://jira.oceanobservatories.org/tasks/browse/OOIION-1774)


lukecampbell:
coi_any build: http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/190

@jamie-cyber1 Here's the patch for one of the issues you saw.

PR #1851, opened 2014-02-27
OOIION-1664 - Adds metadata to ERDDAP Download and Catalog

lukecampbell:
- Parameter Function metadata included in ERDDAP catalog
- Data Product identifier and instrument information included in catalog
- CF compliant attribute names included in catalog

[OOIION-1665](https://jira.oceanobservatories.org/tasks/browse/OOIION-1664)

Note: this pull request is also part of M087


lukecampbell:
@MauriceManning please review

PR #1857, opened 2014-02-27
CIDEVSTC-98 Updates Devices with Site and Deployment attributes upon deployment

brianmckenna:
Updates Devices with Site and Deployment attributes upon deployment and removes when deployment is deactivated.

Device uses Site for geospatial bounds and Deployment for temporal bounds

Upon deactivation, geospatial is returned to the default 0's and temporal bounds is once again copied from the Deployment object


brianmckenna:
@MauriceManning ready for review

mmeisinger:
Can you please rename the functions that are not YML service operations to start with one underscore? We need to distinguish public service ops from "private" class functions

E.g. def _update_device_add_geo_add_temporal

brianmckenna:
@mmeisinger @MauriceManning  should take care of the underscoring

PR #1858, opened 2014-02-27

MauriceManning:
to enable data provenance:
- add name of parameter to related DP instance
- remove hasParentDataProduct from transform/retrieve processes, this is only for parameter function DPDs. TrnasformProcesses and RetrieveProcesses will follow hasInputProduct/hasOutputProduct associations.

@lukecampbell please review
@rajvikram please review


lukecampbell:
Looks like it does what you guys need it to do. I don't have any issues at all.

PR #1863, opened 2014-02-28
PRELOAD - Skips DPD preload for R3 (possibly temporary)

lukecampbell:
I added back the DPD sheets in Preload. This PR will stop ion_loader from preloading the resource, we have a different
way we preload these assets, and we may need to refactor the code if we want to add support back.


mmeisinger:
Totally fine. If needed, R3 DPDs can be preloaded from a DataProcessDefinitionR3 tab.

PR #1867, opened 2014-03-02
TEST FIX: Fix broken test because of device state persister 1s lag

mmeisinger:
Fixes one test that asserts device state but does not wait for the maximum lag of the device state event persister (1s). This PR fixes this by mirroring equal change in master.


mmeisinger:
@jamie-cyber1 please merge

PR #1871, opened 2014-03-03
Bump pyon pointer to include fix for OOIION-1386

jamie-cyber1:


jamie-cyber1:
@mmeisinger please merge when tests pass.

PR #1876, opened 2014-03-05
CIDEVSTC-194: add type and persisted version info

pkediyal:


mmeisinger:
Test builder failed. Merging because it seems unrelated to this PR.

http://buildbot.oceanobservatories.org:8010/builders/coi_pull/builds/672

PR #1880, opened 2014-03-05
CIDEVSTC-49 - Refactors Parameter-based DM service methods to respect IonObjects

lukecampbell:
- Refactors preload for ParameterFunction resources to use a service method instead of independent behavior.
- Adds new service method `create_parameter` which accepts a ParameterContext resource and mimics most of the behavior used in preload.
- Adds new class method `get_coverage_parameter`, creates a coverage-model based ParameterContext object using the ParameterContext IonObject
- Adds new class method `get_coverage_function`, creates a coverage-model based ParameterFunction object using the ParameterFunction IonObject
- Adds new service method `load_parameter_function` which behaves like the preload for parameter functions, this allows service clients to have the same functionality as loading a preloaded resource.
  - Parses information from a row like dictionary
  - Creates an appropriate data process definition
- Changes DM tests to use the new service methods
- Updates types.py to be more robust and not geared only towards preload but for general cases as well.

### DIFF
--- a/ion/processes/bootstrap/ion_loader.py
+++ b/ion/processes/bootstrap/ion_loader.py
@@ -73,7 +73,7 @@ import os
 from udunitspy.udunits2 import UdunitsError
 
 from pyon.core.bootstrap import get_service_registry
-from pyon.core.exception import NotFound
+from pyon.core.exception import NotFound, Conflict
 from pyon.datastore.datastore import DatastoreManager, DataStore
 from pyon.ion.identifier import create_unique_resource_id, create_unique_association_id
 from pyon.ion.resource import get_restype_lcsm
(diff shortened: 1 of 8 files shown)


lukecampbell:
@MauriceManning @mmeisinger please review.

lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/205

MauriceManning:
:+1:

lukecampbell:
After the rebase, another build
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/207

mmeisinger:
Looks good to me on first glance. Refactoring the preload logic to service logic seems done OK.

I'm concerned about runtime of r2deploy and preload. We cannot afford an increase in execution time here. r2deploy is run 1000 times during a test run and needs to be fast.

That's why I implemented the create_parameter_mult function in commit (https://github.com/mmeisinger/coi-services/commit/eb1f93fa9519895afa16c1c00a18c5b6f45ecbde ), which does it all in one bulk operation on the service side. The problem is doing one service call per function/context/dict is probably killing performance because of the resource context to be loaded every single call.

My commit above was a bit of a hack and had strong ties into preload, types etc but it cut the execution time of r2deploy and alpha/beta preload in half. And it worked fine.

Also, it seems the TypesManager is a weird entity in this mix. It's fine for now but I'd like it to be revised/renamed going forward as some kind of parameter management class.

lukecampbell:
Another rebase:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/209

PR #1884, opened 2014-03-05
TEST - Fixes vel3d calibration test

lukecampbell:
Depends on https://github.com/ooici/coi-services/pull/1883


lukecampbell:
here's the test
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/206

PR #1899, opened 2014-03-10
Bump ion-def ptr to turn off validation in launch

jamie-cyber1:


jamie-cyber1:
@mmeisinger please merge.

mmeisinger:
:+1: note that this is further ahead in pyon pointer than master. that's ok

PR #1900, opened 2014-03-10
Bump ion-def ptr to turn off validation during launch

jamie-cyber1:


jamie-cyber1:
@mmeisinger please merge.

mmeisinger:
Cannot merge. The latest ion-definitions needs the latest pyon, because message validators have changed. Otherwise everything will fail.

Good news: pyon is ready too

mmeisinger:
:+1: note, we're not setting to the latest in pyon and ion-definitions. That's ok

PR #1903, opened 2014-03-11
OOIION-1787: Strip extended resource output

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge - submodules are ok

jamie-cyber1:
What about discovery_test?

mmeisinger:
Discovery changed in master. The test does not apply to REL2.0, so there is no change to discovery_test.py included in this PR and there is a difference to master

PR #1907, opened 2014-03-12
Patches the agent controller and adds sample configs

lukecampbell:
- Adds print statements to agent controller (intentional)
- Fixes preload identifier parsing issue
  - The parsing only recognized preload identifiers, not reference designators from the SAF export.
- If the data product isn't active at the time calibrations are set, it attempts to activate the data product.


lukecampbell:
@jamie-cyber1 @mmeisinger 
Please review

Also note: the print statements are intentional, as this is a command line utility. We can remove them after the utility matures a bit more.

jamie-cyber1:
Looks good to me.  Michael?

Where's the pull request for master?

ion/agents/agentctrl.py
```diff
@@ -139,6 +139,7 @@ def _execute_op(self, agent_instance_id, resource_id):
             elif self.op == "config_instance":
                 self._config_instance(agent_instance_id, resource_id)
             elif self.op == "set_calibration":
+                print 'attempting to set calibrations'
```
mmeisinger: Using print is inconsistent with the rest of the tool and should be avoided. I understand why but we do not want this inconsistency. Just change the log level of the agentctrl tool to DEBUG

ion/agents/agentctrl.py
```diff
@@ -320,12 +321,21 @@ def _set_calibration(self, agent_instance_id, resource_id):
         res_obj = self.rr.read(resource_id)
 
         # Device has no reference designator - but use preload ID as reference designator
-        alt_ids = [aid[4:] for aid in res_obj.alt_ids if aid.startswith("PRE:ID")]
+        alt_ids = []
+        for aid in res_obj.alt_ids:
+            if aid.startswith('PRE:') and aid.endswith('_ID'):
+                alt_ids.append(aid[4:-3])
+            elif aid.startswith('PRE:ID'):
```
mmeisinger, line 327: There should be no preload ID starting with PRE:ID, so this conditions should be removed. I would not know where this comes from and if it is there it needs to be removed.

lukecampbell, line 327: All of the instrument devices in preload start with PRE:ID. The only ones that do not are the SAF instruments which are appended with _ID. This was the core of the bug and why it didn't work. If I change it then preloaded instruments won't work.

ion/agents/agentctrl.py
lukecampbell: I'll remove the prints when I have a moment to change the code. I really would like to see prints in these command line utilities. The developers that are running them always ask me to add more information to the output when running them, and it's always because the logging levels are not correct or something to do with logging not being enabled (or explicitly disabled). My solution was to add prints, guaranteed stdout delivery.

ion/agents/agentctrl.py
mmeisinger, line 327: I see. The PRE:ID are for the manually entered bench devices, not for the OOI preloaded resources. It's very unstable to use them but ok for now.

ion/agents/agentctrl.py
mmeisinger: I see the point here but it needs to be consistent across the tool. We had prints in bin/pycc but Jonathan changed them to log. In the end it's the integration team's call and then need to be done consistently. If there is log there should not be print

mmeisinger:
I'm ok to merge at this point

jamie-cyber1:
We still need this master Luke.

ion/agents/agentctrl.py
lukecampbell, line 327: When I tested this out initially I used the devices defined by Beta (namely the simulator) which is why I had it the other way. I didn't realize the ooi preloaded assets used _ID.

PR #1908, opened 2014-03-12
STCDEV M101 in progress.

edwardhunter:
Change import order so test classes are first, OT expanded (minor adjustment so tests run without pycc).
@jamie-cyber1 review.


edwardhunter:
@jamie-cyber1 review.

PR #1923, opened 2014-03-13
CIDEVSTC-49 - New service method for adding a parameter to a data product

lukecampbell:
Only adding it to a dataset doesn't update the stream definition or the datasets catalog, this method provides clients to add a parameter to a data product and have it propagate through the system correctly


lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/219

lukecampbell:
@jamie-cyber1 Ready to merge

PR #1930, opened 2014-03-17
OOIION-1749 Ensure state object is flushed on publication

wfrench:
@edwardhunter - Can you please review.

I slightly diverged on when we flush in the ASP.


mmeisinger:
Please check whether you need to flush after this line, immediately after an DSA agent start with prior state:

https://github.com/ooici/coi-services/blob/master/ion/agents/data/dataset_agent.py#L216

Otherwise there is danger that this updated state will not get flushed right away and then lost at a system restart.

PR #1934, opened 2014-03-18
OOIION-869: Enhanced UNS find_events and polished interface

mmeisinger:


mmeisinger:
@jamie-cyber1 please review

jamie-cyber1:
ion-def merged.  Need pointer updates.

mmeisinger:
@jamie-cyber1 ion-def updated, merge when ready

ion/services/dm/presentation/user_notification_service.py
```diff
@@ -897,4 +695,120 @@ def check_publish_event_policy(self, process, message, headers):
...
+            origin_type = origin_type,
+            sub_type = sub_type,
+            description = description,
+            **event_attrs
+            )
+        log.info("Event published: %s", event)
+
+        return event
+
+    def find_events(self, origin='', type='', min_datetime='', max_datetime='', limit=-1,
```
jamie-cyber1, line 733: Why is limit -1 instead of 0?

mmeisinger, line 733: This is code that existed before. I just moved it around. If there was no error before, we should just leave it. When I initially did this I remember an issue. I'm not sure if I can reproduce it but it was along the following lines:
- If you make a service call, the service client automatically enters the YML default (0 in this case)
- If you set an explicit value (other than 0) it will be used
- The meaning of -1 is no limit
- The meaning of 0 is apply the container defined limit

I think we are good and this can be mergee

jamie-cyber1:
Michael, need to update pointer.

mmeisinger:
Pointer updated

PR #1948, opened 2014-03-21
CIDEVSTC-290 Data Product Metadata and Catalog Management

lukecampbell:
- Adopts a template based system to allow a more granular externalization for data product metadata
- Provides individual CRUD operators for catalog entries of data products
- Adds consistency for the catalog by keeping a working copy in the Directory.
- Changes to the catalog are atomic in nature now, so it's less prone to failures because of us adding it in the Directory.
- Considerably more metadata are available in ERDDAP now.

### DIFF
--- a/ion/services/sa/product/data_product_management_service.py
+++ b/ion/services/sa/product/data_product_management_service.py
@@ -11,6 +11,7 @@ from pyon.util.ion_time import IonTime
 from pyon.ion.resource import ExtendedResourceContainer
 from pyon.event.event import EventPublisher
 from pyon.util.arg_check import validate_is_instance, validate_is_not_none, validate_false, validate_true
+from pyon.net.endpoint import RPCClient
 
 from ion.services.dm.utility.granule_utils import RecordDictionaryTool
 from ion.util.enhanced_resource_registry_client import EnhancedResourceRegistryClient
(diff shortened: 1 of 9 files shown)


lukecampbell:
@MauriceManning @mmeisinger Ready for review please.

lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/232

lukecampbell:
Oh yeah forgot, another big change about this is that PyDAP now accesses the container, it's no longer separable (without additional work). It uses the resource registry to identify associated datasets, this was actually in-preparation for M193.

ion/services/sa/product/data_product_management_service.py
```diff
@@ -227,9 +218,21 @@ def update_data_product(self, data_product=None):
...
         self.RR2.update(data_product, RT.DataProduct)
+        if self._metadata_changed(original, data_product):
+            self.update_catalog_entry(data_product._id)
+
+    def _metadata_changed(self, original_dp, new_dp):
+        from ion.processes.data.registration.registration_process import RegistrationProcess
+        for field in RegistrationProcess.catalog_metadata:
+            if hasattr(original_dp, field) and getattr(original_dp, field) != getattr(new_dp, field):
+                return True
+        return False
```
lukecampbell, line 232: I wanted to point this out, we'll only update a catalog entry through update_data_product if the metadata fields have changed. I'm trying to reduce the frequency at which we could potentially update the catalog.

mmeisinger:
Looks good to me (not knowing the details). Good use of the directory.

lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/234

PR #1949, opened 2014-03-24
M114 Coverage Model Integration

caseybryant:
Bump coverage model version.


caseybryant:
@jamie-cyber1 please review and merge.

jamie-cyber1:
The smoke tests failed.

caseybryant:
I ran smoke and unit on two different repositories before issuing the pull request.  I reran the test again locally, and cannot reproduce the error.  The logs indicate that the data being submitted to coverage-model is invalid.

I did a force build on buildbot to rerun the test and it was successful - http://buildbot.oceanobservatories.org:8010/builders/coi_pull/builds/788

I don't like that.  Any idea (resource sharing, etc.) why the first build might have failed?

jamie-cyber1:
We are sharing one postgres instance.  Do you use any hardcoded db resources that might cause a fail when parallel builds are running?

caseybryant:
I shouldn't be.  I access Postgres from pyon so unless pyon does, that should be the problem.

PR #1953, opened 2014-03-24
M114 Coverage Model Integration

caseybryant:
Skip tests that corrupt hdf metadata files.  HDF is no longer used the backing store for metadata.


caseybryant:
@jamie-cyber1 These changes will fix two of the failures with coi-nightly.

PR #1954, opened 2014-03-25
CIDEVSTC-302 - Adds escape handling for sensitive characters in XML

lukecampbell:
- Certain characters need to be properly escaped before going into
  datasets.xml


jamie-cyber1:
:+1:

lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/235

PR #1962, opened 2014-03-26
updates default test values for skipping available tests

birdage:
update default eoi service references


birdage:
@jamie-cyber1 ready for merge, should fix the build issue

PR #1965, opened 2014-03-27
Bump ion-defs

lukecampbell:


lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/236

PR #1968, opened 2014-03-28
Adds utility to find the reference designator for a data product

lukecampbell:


lukecampbell:
@mmeisinger @MauriceManning please review.

mmeisinger:
Looks good but rests on the assumption that there is always only one device per data product. That's probably fine for now.

lukecampbell:
Good to know.

PR #1972, opened 2014-03-31
Skip tests in cei mode

jamie-cyber1:


jamie-cyber1:
@mmeisinger please review and merge.

PR #1979, opened 2014-04-03
CIDEVSTC-287 - M088 - QC tables upload (CSV file)

brianmckenna:
@MauriceManning @lukecampbell

adds integration test and skips blank lines in CSV

### DIFF
--- a/ion/services/dm/test/test_uploads.py
+++ b/ion/services/dm/test/test_uploads.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python
+
+__author__ = 'Brian McKenna <bmckenna@asascience.com>'
+
+import json
+import os
+import tempfile
+import time
+from nose.plugins.attrib import attr
+from webtest import TestApp
+from ion.services.coi.service_gateway_service import service_gateway_app
+from interface.services.coi.iservice_gateway_service import ServiceGatewayServiceClient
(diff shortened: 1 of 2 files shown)


ion/services/dm/test/test_uploads.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # read the FileUploadContext (async) return
+        #gateway_response = data.get('GatewayResponse', None)
+        #fuc_id = gateway_response.get('fuc_id', None) # CAUTION this is unicode
+        #fuc = self.object_store.read(str(fuc_id))
+
+        # check the QC table stored in object_store
+        REFDES01 = self.object_store.read('REFDES01')
+        RD01DP01 = REFDES01.get('RD01DP01', None)
+        assert RD01DP01 == {
```
lukecampbell: could you change this line to be 

``` python
self.assertEquals(RD01DP01, {...
```

py.test would work well with this syntax and show what parts aren't equal if there are incorrect values. However, unittest doesn't flag it, it just shows the test failing (or maybe even error).

lukecampbell:
looks good otherwise :+1:

lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/237

PR #1985, opened 2014-04-04
CIDEVSTC-183 Integrate and Validate with Production System

MauriceManning:
add updates to script; logging, call ion-ux web svr, error checking, data folder naming

@mmeisinger please review

@rajvikram please review

### DIFF
--- a/ion/ext/tools/data_product_monitor.py
+++ b/ion/ext/tools/data_product_monitor.py
@@ -2,42 +2,78 @@ import sys
 import os.path
 import json, time
 import urllib, urllib2
+import inspect
+import logging
+import argparse
 
-# $ pip install requests
-import requests
+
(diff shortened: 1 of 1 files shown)


mmeisinger:
:+1: Looks really nice and to the point

Do you have somewhat noticeable a way of reacting to exceptions from the get_data_product_updates call, e.g. if the user specified bad DP id or something else went awry?

Somewhere should be a snippet of documentation for end users for how to download the script, create an environment with necessary dependencies (if any), which constants in the script or config file to change and how to run.

MauriceManning:
Invalid ids or other issues are logged as errors or warnings currently. Will work with Tim and Jamie next week to determine how users to download the script. Will add README section to the top of the script today.

PR #1987, opened 2014-04-04
Bumped pyon

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

This one jumps 3 instead of 2 PRs. We missed one before, but it only fixes some tests. Should be ok

PR #1993, opened 2014-04-04
CIDEVSTC-288 - QC Processing

lukecampbell:
The first iteration of the QC Processor. 
- Doesn't store data until the CM is ready but it gets the data as close as possible to ready
- Still needs to apply windows and RD lookups for multi-deployment

It actually works!

### DIFF
--- a/ion/processes/bootstrap/ion_loader.py
+++ b/ion/processes/bootstrap/ion_loader.py
@@ -1939,35 +1939,6 @@ Reason: %s
                         context.lookup_value = name
                         context.document_key = lookup_value
             
-            qc_map = {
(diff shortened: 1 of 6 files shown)


lukecampbell:
@MauriceManning ready for your review.
There's no tests and it's no where near ready for integration (because of M088's dependencies) but worst-comes-to-worst it runs and we can integrate it easily if it becomes imperative.

lukecampbell:
I'm gonna put back in the prelaod stuff, I think it's premature to make that jump.

lukecampbell:
There, now we can safely merge it.

MauriceManning:
:+1:  good start.

PR #2003, opened 2014-04-10
Clones the Parameter Dictionary for every New Data Product

lukecampbell:
- Cloned Parameter Dictionaries have the name DataProduct.name + _ + DataProduct._id
- Preload is a bit slower because it doesn't mimic the service yet which leads to cache misses for the parameter dictionaries in the loading process
  - For now, I'd rather preload go slower than have consistency issues
- Removes DPMS Unit test. The test serves little more than a syntax check and it's now deprecated. Frankly, I'm surprised it worked because of the refactor we did a few months ago. When we split up the creation into multiple methods that the test should have broke, I think.

### DIFF
--- a/ion/processes/bootstrap/ion_loader.py
+++ b/ion/processes/bootstrap/ion_loader.py
@@ -3129,6 +3129,11 @@ Reason: %s
 
         for assoc in assocs_filtered:
             pdict = self._get_resource_obj(assoc.s)
+            # If the parameter dictionary was created by the service logic then it won't be cached in the loader
+            # In this case, we need to pull it from the resource registry
+            # TODO: Update the preload logic to mimic the service way
+            if pdict is None:
+                pdict = self.container.resource_registry.read(assoc.s)
             pdef = self._get_resource_obj(assoc.o, True)
             if pdef is None:
                 log.warn("Ignoring ParameterContext %s - not found", assoc.o)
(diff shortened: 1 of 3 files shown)


lukecampbell:
@mmeisinger @MauriceManning Please review

lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/241

mmeisinger:
Looks ok, but it is a somewhat dangerous approach if I understand it correctly. You clone on assign. Assign now has a different side effect than what you would expect. 

Why can't we create the individual PDICTs in preload and assign without cloning? This trick may bit us in the behind later.

lukecampbell:
We can create the individual pdicts inside of the _load_DataProduct_OOI, I just haven't gotten to it yet, and I'm not sure how it would look. Preload is very different from the services, it deals in bulk.

I'm just trying to get the concept through and agreed on.

Maybe I'll break it up before assign, I just thought assign was a nice place, there is already an unintuitive side-effect which is creating the streams (there's a todo and a comment above it). Maybe the best solution is to change _assign_X_to_Y to something that's a bit more intuitive like setup_stream_and_parameters

lukecampbell:
I just spent some time debugging the failure.

This isn't going to work...

The instrument agent determines the right stream by looking at the parameter dictionaries and working its way backwards...

PR #2006, opened 2014-04-10
M088 - type translation of QC uploads

brianmckenna:
type translation before insertion into object_store (was strings, now specific types)

@lukecampbell check this is what you expect (see test asserts)
@MauriceManning


lukecampbell:
Yeah looks good, thanks.

MauriceManning:
:+1:

PR #2009, opened 2014-04-14
OOIION-1835: Fix roles UI issues (also fixes OOIION-1831)

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #2011, opened 2014-04-14
OOIION-1837: Port get names for events code back to R2

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge (this affects R2 only. R3 is ok)

PR #2012, opened 2014-04-14
M088 - QC tables - download via DatasetManagementService

brianmckenna:
@MauriceManning
@lukecampbell

requires https://github.com/ooici/ion-definitions/pull/548

adds read_object to dataset management service and test to read QC tables from object store through DMS


lukecampbell:
Can you bump the submodule, after that I'll merge

PR #2014, opened 2014-04-14
M058 - Make changes to the resource parser to allow RPS calls from a process

birdage:


birdage:
@jamie-cyber1 ready for merge
@lukecampbell for info

lukecampbell:
Looks good!

birdage:
for info this is related to the additional tasks, M058:T33:A01

ion/services/eoi/table_loader.py
```diff
@@ -21,58 +23,68 @@
...
     """
-    def __init__(self):
+    def on_start(self):
+
+        SimpleProcess.on_start(self)
+
         self.using_eoi_services = CFG.get_safe('eoi.meta.use_eoi_services', False)
+        if not self.using_eoi_services:
+            # place holder
+            raise BadRequest("well deal with it later...")
```
jamie-cyber1, line 37: @mmeisinger: with BadRequest, the process would fail and get restarted indefinitely by PD with default restart_mode of ABNORMAL? What's the best way to handle this?

birdage, line 37: @lukecampbell can you remember the rational as to why we did this?

lukecampbell, line 37: lol oops.  

That should be something like 

``` python
raise BadRequest("table_loader process shouldn't be launched without eoi.meta.use_eoi_services set to True")
```

birdage, line 37: we might want to remove the bad request and skip the setup if the user has set the "using_eoi_services" bool to false in the yml file.

lukecampbell, line 37: This would only happen in a very weird accidental situation. The bootstrap_process_dispatcher module only launches this process if CFG allows. But, in a very weird situation, I suppose an operator could try to manually launch this process. We don't want this process running unless this CFG variable is set, I think.

ion/processes/bootstrap/plugins/bootstrap_process_dispatcher.py
```diff
@@ -39,7 +40,24 @@ def pydap_server(self, process, config):
...
 
+    def importer_server(self, process, config):
+        res, meta = self.resource_registry.find_resources(name='table_loader', restype=RT.ProcessDefinition)
+        if len(res):
+            return
+
+        table_loader_module = config.get_safe('bootstrap.processes.table_loader.module', 'ion.services.eoi.table_loader')
+        table_loader_class  = config.get_safe('bootstrap.processes.table_loader.class', 'ResourceParser')
 
+        use_table_loader = config.get_safe('eoi.meta.use_eoi_services', False)
```
birdage, line 51: @lukecampbell should only start an instance if its available

ion/services/eoi/table_loader.py
birdage, line 37: @lukecampbell that is correct, it should only be started if the bootstrap process is valid, and if we want it to run.

ion/processes/bootstrap/plugins/bootstrap_process_dispatcher.py
lukecampbell, line 51: Yes, however, it's absolutely possible to launch this process manually. We should try to prohibit that to a certain extent.

birdage, line 51: @lukecampbell well discuss and ill make a change.

jamie-cyber1:
This design seems to be using one table_loader singleton process. And then the many dataset management services spawned would be bottlenecked at issuing rpc requests to that table_loader.  Is that right?

lukecampbell:
Jamie, that's correct. table_loader is a singleton. Dataset management could potentially be bottlenecked if the table_loader takes too long.

ion/services/dm/inventory/dataset_management_service.py
```diff
@@ -754,3 +741,70 @@ def _merge_contexts(cls, contexts, temporal):
                 pdict.add_context(context)
         return pdict
 
+    def create_single_resource(self,dataset_id, param_dict):
```
mmeisinger, line 744: Please make sure that functions in as service that are not part of the public interface start with underscore, to distinguish them from public service operations

ion/services/eoi/table_loader.py
```diff
@@ -21,58 +23,68 @@
...
+            log.debug("TableLoader:NOT using geoservices...")
 
-            
-            if self.postgres_db_available and self.importer_service_available:
-                self.use_geo_services = True
-                log.debug("TableLoader:Using geoservices...")
-            else:
-                log.debug("TableLoader:NOT using geoservices...")
+        #generate rpc server after everything has been setup
+        self._rpc_server = self.container.proc_manager._create_listening_endpoint(from_name=self.id, process=self)
```
mmeisinger, line 83: I don't like the fact that we created a new ION process for this, but given we do in this code, why don't we use StandaloneProcess, which automatically creates a self.id RPC listening endpoint? Why SimpleProcess and the manual endpoint?

ion/services/dm/inventory/dataset_management_service.py
```diff
@@ -754,3 +741,70 @@ def _merge_contexts(cls, contexts, temporal):
...
+        '''
+
+        # See if we have it already before hitting RPC
+        if self.geos_available:
+            return True
+        try:
+            pid = self._get_table_loader_pid()
+        except BadRequest as e:
+            if "No Table Loader Found..." in e.message:
+                log.debug("No Table Loader Found")
```
mmeisinger, line 784: A condition like this should probably be log.warn

ion/services/eoi/table_loader.py
lukecampbell, line 83: We can move it to standalone process.

mmeisinger:
Creating a new RPC process for this purpose was not part of the agreement last time we talked. Every RPC process has an architectural cost. This one is non-standard and does not even have a YML interface and as a singleton will cause problems in the future. I'm a bit disappointed that we don't discuss architectural changes of this magnitude beforehand.

I'm not yet fully positive how I stand wrt to the current code proposal. Why was it not possible to put the table import into the dataset management service, and only do basic initialization in the bootstrapper?

I'd like to see justification why we need a new RPC process in the system.

lukecampbell:
I believe the reasoning is because table_loader is a singleton and dataset management is not. We don't have transactional safety inherent in the services: one client can update a resource while another deletes one and inconsistencies can arise, I think. To mitigate the risk of this happening with this external integration and interface with geoserver, we opted for a singleton. 

The only way I, personally, saw to facilitate a single interface to geoserver in a distributed environment was if we created a singleton process and used RPC. I agree that funneling the RPC calls from a multi-instance service like DSM to a singleton has the potential for bottlenecks. However, I think that the execution time for these is very minimal.

lukecampbell:
Back to the drawing board.

mmeisinger:
There is no inherent danger in concurrency conflicts. If one updates and the other deletes then there are 2 possible outcomes, no matter at which level you try to serialize. Each caller will need to be prepared for dealing with concurrency problems, which can happen anywhere in the system. Why would it be critical for a system that mirrors datasets to an externalization server? Mirroring should be idempotent, e.g. when run again (given the same resources) it should produce the same result.

Dataset management service has multiple workers for reasons of load balance. If it takes a few seconds to register a dataset with GeoServer postgres, it's the perfect place for this. There will be plenty other workers that deal with other service requests in parallel. Funneling the long latency code into a singleton is exactly the opposite of what services should do. Most of the time, service business logic (e.g. a RR read or update) does not fully justify the cost associated with the RPC service call. In the cases where business logic costs time, it must be in a parallelizable service. If there are concurrency issues, they need to be dealt with on the higher level.

I suggest making the TableLoader a class (as it was), used by the service workers. Put the one-time database initialization logic into the bootstrapper. Have the service instances maintain postgres connections to the GeoServer postgres. Log warnings and let the operator deal with inconsistencies.

lukecampbell:
Yeah, I think that's what we're going to do.

PR #2024, opened 2014-04-21
OOIION-1565 Refactor deployment activation logic

MauriceManning:
fix issues in deployment preload and make all ref designators unique

@jamie-cyber1 please review
@mmeisinger please review


mmeisinger:
Build failed. Should this be a R2 PR? Will preload be backward compatible?

PR #2025, opened 2014-04-21
OOIION-1844: Expand length of postgres dir key column

jamie-cyber1:


jamie-cyber1:
@mmeisinger please merge.

PR #2030, opened 2014-04-23
M058 : Update bootstrap to include eoi services

birdage:
The ResourceParser object is instantiated once at the bootstrap plugin level to reset the Geoserver instance(s). 
Each DatasetManagementService instance has a ResourceParser instantiated in it and interfaces with it on client requests that create or delete a dataset resource.

### DIFF
--- a/ion/processes/bootstrap/plugins/bootstrap_eoi.py
+++ b/ion/processes/bootstrap/plugins/bootstrap_eoi.py
@@ -15,7 +15,7 @@ class BootstrapEOI(BootstrapPlugin):
         """
         EOI BootstrapPlugin
 
-        Creating transform workers happens here...
+        Resets the geoserver datastore... 
         """
         r = ResourceParser()
         r.init()
--- a/ion/processes/bootstrap/plugins/bootstrap_eoi.py
+++ b/ion/processes/bootstrap/plugins/bootstrap_eoi.py
@@ -19,14 +19,14 @@ class BootstrapEOI(BootstrapPlugin):
 
         Resets the geoserver datastore... 
         """
-        r = ResourceParser()
-        try:
-            r.init()
-            r.reset()
-        except BadRequest as e:
-            if 'Eoi services not enabled' in e.message:
-                log.warning("EOI Services are not enabled")
-        except:
-            log.error("EOI Services are disabled", exc_info=True)
+        using_eoi_services = config.get_safe('eoi.meta.use_eoi_services', False)
+        if using_eoi_services:
+            try:
+                r = ResourceParser()
+                r.reset()
+            except Exception:
+                log.error("Error resetting EOI Services")
+        else:
+            log.info("EOI Services are not enabled")
         
         
--- a/ion/processes/bootstrap/plugins/bootstrap_eoi.py
+++ b/ion/processes/bootstrap/plugins/bootstrap_eoi.py
@@ -18,6 +18,10 @@ class BootstrapEOI(BootstrapPlugin):
         Resets the geoserver datastore... 
         """
         r = ResourceParser()
-        r.init()
-        r.reset()
+        try:
+            r.init()
+            r.reset()
+        except Exception, e:
+            raise e
+        
         
--- a/ion/processes/bootstrap/plugins/bootstrap_eoi.py
+++ b/ion/processes/bootstrap/plugins/bootstrap_eoi.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python
+
+from ion.core.bootstrap_process import BootstrapPlugin
+from pyon.public import Container
+from interface.objects import IngestionQueue
+from interface.services.dm.iingestion_management_service import IngestionManagementServiceProcessClient
+from ion.services.eoi.table_loader import ResourceParser
+
+class BootstrapEOI(BootstrapPlugin):
+    """
+    Bootstrap process for EOI management.
+    """
+
+    def on_initial_bootstrap(self, process, config, **kwargs):
+        """
+        EOI BootstrapPlugin
+
+        Creating transform workers happens here...
+        """
+        r = ResourceParser()
+        r.init()
+        r.reset()
+        
--- a/ion/processes/bootstrap/plugins/bootstrap_eoi.py
+++ b/ion/processes/bootstrap/plugins/bootstrap_eoi.py
@@ -6,6 +6,7 @@ from interface.objects import IngestionQueue
 from interface.services.dm.iingestion_management_service import IngestionManagementServiceProcessClient
 from ion.services.eoi.table_loader import ResourceParser
 from pyon.util.log import log
+from pyon.core.exception import BadRequest
 
 class BootstrapEOI(BootstrapPlugin):
     """
@@ -22,7 +23,10 @@ class BootstrapEOI(BootstrapPlugin):
         try:
             r.init()
             r.reset()
-        except Exception, e:
-            log.error("Failed to reset and initialize eoi services", exc_info=True)
+        except BadRequest as e:
+            if 'Eoi services not enabled' in e.message:
+                log.warning("EOI Services are not enabled")
+        except:
+            log.error("EOI Services are disabled", exc_info=True)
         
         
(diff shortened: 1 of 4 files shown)


birdage:
@MauriceManning 
@mmeisinger 
@jamie-cyber1 
@lukecampbell 
Please review

birdage:
requires https://github.com/ooici/ion-definitions/pull/557

lukecampbell:
Just FYI, the build failed because it doesn't include the submodule, we'll bump the submodule once it's been merged.

ion/processes/bootstrap/plugins/bootstrap_eoi.py
```diff
@@ -0,0 +1,23 @@
...
+class BootstrapEOI(BootstrapPlugin):
+    """
+    Bootstrap process for EOI management.
+    """
+
+    def on_initial_bootstrap(self, process, config, **kwargs):
+        """
+        EOI BootstrapPlugin
+
+        Creating transform workers happens here...
```
lukecampbell: When we do the bump, we probably need to change this description.

mmeisinger:
Looks good to me. Much better.

jamie-cyber1:
A lot of these debug statements on 'geo service or postgres not being available' should probably be logged at level warning instead?

lukecampbell:
Ok, we're going to chnge the boot level in ion-defs, drop the prints change the docstring then bump the submodule.

birdage:
@lukecampbell @mmeisinger 
Please review, waiting on merge of https://github.com/ooici/ion-definitions/pull/557 before i can bump the submodule

ion/processes/bootstrap/plugins/bootstrap_eoi.py
```diff
@@ -0,0 +1,23 @@
...
+
+class BootstrapEOI(BootstrapPlugin):
+    """
+    Bootstrap process for EOI management.
+    """
+
+    def on_initial_bootstrap(self, process, config, **kwargs):
+        """
+        EOI BootstrapPlugin
+
```
mmeisinger, line 19: Do you have to put an if enabled statement here or should this code be run in any condition?

mmeisinger:
ion-definitions have been merged, please update.

See my comment about the if-statement in the bootstrapper.

Other than that this is ready to go

ion/processes/bootstrap/plugins/bootstrap_eoi.py
```diff
@@ -0,0 +1,23 @@
...
+        Resets the geoserver datastore... 
+        """
+        r = ResourceParser()
+        r.init()
+        r.reset()
```
birdage: the reset has a verification check that geoservices are available see https://github.com/birdage/coi-services/blob/bs-rb/ion/services/eoi/table_loader.py#L126

lukecampbell: Could we throw in a try except around this and catch BadRequest from https://github.com/birdage/coi-services/blob/bs-rb/ion/services/eoi/table_loader.py#L41

I didn't think about the situation but if EOI services aren't enabled we can't actually start a container now.

lukecampbell: Also add

``` python
from pyon.core.exceptions import BadRequest
```

 to the top of table_loader.py it doesn't know about BadRequest.

mmeisinger: Yes this was the one I suspected. 
I'd rather do a if-else but anything works

birdage: i will make that change now

birdage:
@mmeisinger @lukecampbell 
Please Review

ion/processes/bootstrap/plugins/bootstrap_eoi.py
```diff
@@ -0,0 +1,27 @@
...
+        """
+        EOI BootstrapPlugin
+
+        Resets the geoserver datastore... 
+        """
+        r = ResourceParser()
+        try:
+            r.init()
+            r.reset()
+        except Exception, e:
```
mmeisinger: What does this do?
Also catching any kind of exception at this critical juncture of the system (bootstrap) is poor form. Anything may go wrong here and you should not mask this.
I see that you reraise, so this try-except has no function

lukecampbell:
I'll merge once test finishes.

ion/processes/bootstrap/plugins/bootstrap_eoi.py
birdage: updated in latest commit

ion/processes/bootstrap/plugins/bootstrap_eoi.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+
+    def on_initial_bootstrap(self, process, config, **kwargs):
+        """
+        EOI BootstrapPlugin
+
+        Resets the geoserver datastore... 
+        """
+        r = ResourceParser()
+        try:
```
mmeisinger: if r. get_eoi_service_available():
   r.init()
   r.reset()
else:
  log.info("EOI Services are not enabled")

ion/processes/bootstrap/plugins/bootstrap_eoi.py
```diff
@@ -0,0 +1,32 @@
...
+            r.init()
+            r.reset()
+        except BadRequest as e:
+            if 'Eoi services not enabled' in e.message:
+                log.warning("EOI Services are not enabled")
+        except:
+            log.error("EOI Services are disabled", exc_info=True)
```
mmeisinger: log.exception("Error initializing EOI services")

ion/processes/bootstrap/plugins/bootstrap_eoi.py
```diff
@@ -0,0 +1,32 @@
...
+        Resets the geoserver datastore... 
+        """
+        r = ResourceParser()
+        try:
+            r.init()
+            r.reset()
+        except BadRequest as e:
+            if 'Eoi services not enabled' in e.message:
+                log.warning("EOI Services are not enabled")
+        except:
```
mmeisinger: except without any type is illegal (see code conventions).

at least do except Exception:

lukecampbell: I think we're screwing up in a lot more places than just this PR then.
There are 128 occurrences of this exact pattern in our current code base.
Here is a comprehensive list
https://gist.github.com/lukecampbell/11299360

I'm guessing the code conventions were published after these code modules were committed. 

I'll change the code shortly.

ion/processes/bootstrap/plugins/bootstrap_eoi.py
lukecampbell: that method isn't available until init has been called
https://github.com/birdage/coi-services/blob/bs-rb/ion/services/eoi/table_loader.py#L65
https://github.com/birdage/coi-services/blob/bs-rb/ion/services/eoi/table_loader.py#L87-L91

mmeisinger: Make it available then after __init__
Relying on an exception and checking the e.message seems not right for new code that you have full control over

ion/processes/bootstrap/plugins/bootstrap_eoi.py
lukecampbell: May I inquire as to why we should use exception here instead of error with an exception print out?
We do it in a lot of other places
https://gist.github.com/lukecampbell/11299523

ion/processes/bootstrap/plugins/bootstrap_eoi.py
mmeisinger: The only means for me to enforce conventions is through code reviews and PRs. Especially for developers new on the project how else would they learn? The fact that the existing code has bad patterns is a sad fact of life and can only slowly be overcome but is not an excuse to continue bad practices. I can only work on educating the use of better patterns, but it's rare that devs go back fix their code once it's in use. Do it upfront.

ion/processes/bootstrap/plugins/bootstrap_eoi.py
mmeisinger: This was a rather small hint, mainly because the error message was wrong. This branch was NOT concerned about EOI disabled, but any other error.

log.exception() is a shorthand for log.error("msg", exc_info=True) and more in line with the intent.
https://docs.python.org/2/library/logging.html#logging.Logger.exception

PR #2031, opened 2014-04-23
CIDEVSTC-288 - Event handling in the QC Process

lukecampbell:
- Refactors preload a bit to move all the qc handling into a function that's called by load parameter
- Adds an event subscriber to QC Processor to listen for incoming events and then append them to a queue
- At the conclusion of the qc processing loop, each event in the event queue is handled.
  - The events will typically result in QC being re-evaluated for larger sets of data.

### DIFF
--- a/ion/processes/bootstrap/ion_loader.py
+++ b/ion/processes/bootstrap/ion_loader.py
@@ -1965,35 +1965,7 @@ Reason: %s
                         context.lookup_value = name
                         context.document_key = lookup_value
             
-            qc_map = {
-                    'Global Range Test (GLBLRNG) QC'                         : 'glblrng_qc',
-                    'Stuck Value Test (STUCKVL) QC'                          : 'stuckvl_qc',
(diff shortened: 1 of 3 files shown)


lukecampbell:
@MauriceManning 
@mmeisinger 
Ready for review and discussion

mmeisinger:
Refactoring looks good. The event type is probably not right. Is it really ParameterQCEvent? Do we need to refine the event type hierarchy wrt QC events? There are TONS of ParameterQCEvent in the system.

I don't see high level try-except in your basic thread loops. Make sure these cannot die

MauriceManning:
We might want to create the QCProcess event under DataProcessEvent?

lukecampbell:
The event type doesn't exist yet, I just used QCParameterEvent as a place holder until I iron out all the cases which require events from the requirements. There's a few.

lukecampbell:
Just a point of clarification, this PR doesn't affect the system in any way. I'm just trying to get code merged in while also stimulating understanding about what's happening at this level through discussions in PRs.

PR #2038, opened 2014-04-24
CIDEVMI-1178 Updates ion-functions

cwingard:
Corrects volt to millivolt conversion for calculation of OPTPARW from the PARAD-K on the WFP.


cwingard:
@jamie-cyber1 Ready for review and merge

PR #2047, opened 2014-04-29
un-skip platform-agent/mission-execution tests

carueda:


carueda:
@edwardhunter  review/merge

Note the TODO about pending adjustment in PA to check child state before attempting sending command.

My local testing looks like:

```
bin/nosetests -sv --nologcapture  --with-pycc ion/agents/platform/test/test_mission_manager.py
...
----------------------------------------------------------------------
Ran 2 tests in 222.158s

OK
```

(so, per Bob's recent adjustment in the test mission file, this went from ~ 600 secs down to 222 secs)

PR #2048, opened 2014-04-29
OOIION-1823: Correct OOI preload of generated resource attributes

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #2052, opened 2014-05-01
Bump pyon

lukecampbell:
Adds in environment variable check before executing force_clean in PyonTestCase.


jamie-cyber1:
This was a great catch!

PR #2055, opened 2014-05-02
M193 - SiteDataProducts

kwilcox:
- Attempt to create SiteDataProducts in preload.  Emphasis on "Attempt".  The test that preloads alpha to test SiteDataProducts fails.  The tests in which the resources are manually created all pass.
- Creates the "withinDeployment" association when a deployment is activated. This association is never deleted.
- Creates the "hasPrimaryDeployment" association when a deployment is activated. This association is deleted on deactivation.
- Requires new preload column 'stream_configurations' on the [Platform|Instrument]Site sheets.

### DIFF
--- a/ion/processes/bootstrap/ion_loader.py
+++ b/ion/processes/bootstrap/ion_loader.py
@@ -92,7 +92,7 @@ from coverage_model import NumexprFunction, PythonFunction, QuantityType, Parame
 
 from interface import objects
 from interface.objects import StreamAlertType, PortTypeEnum, StreamConfigurationType, ParameterFunction as ParameterFunctionResource
-from interface.objects import DataProcessDefinition, DataProcessTypeEnum, StreamConfiguration
+from interface.objects import DataProcessDefinition, DataProcessTypeEnum, StreamConfiguration, StreamDefinition, DataProduct, DataProductTypeEnum
 
 from ooi.timer import Accumulator, Timer
 stats = Accumulator(persist=True)
@@ -1649,11 +1649,17 @@ class IONLoader(ImmediateProcess):
     def _load_PlatformSite(self, row):
         constraints = self._get_constraints(row, type='PlatformSite')
         coordinate_name = row['coordinate_system']
+        coordinate_reference_system = self.resource_ids[coordinate_name] if coordinate_name else None
+        # Get StreamConfiguration objects by the ID in stream_configurations row
+        stream_configurations = []
+        if "stream_configurations" in row and row['stream_configurations']:
+            stream_configurations = map(lambda x: self.stream_config[x], row['stream_configurations'].split(","))
 
         res_id = self._basic_resource_create(row, "PlatformSite", "ps/",
             "observatory_management", "create_platform_site",
             constraints=constraints, constraint_field='constraint_list',
-            set_attributes=dict(coordinate_reference_system=self.resource_ids[coordinate_name]) if coordinate_name else None,
+            set_attributes=dict(coordinate_reference_system=coordinate_reference_system,
+                                stream_configurations=stream_configurations),
             support_bulk=True)
 
         if self.bulk:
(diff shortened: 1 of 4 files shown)


kwilcox:
PS this is only for review @edwardhunter @mmeisinger @MauriceManning

MauriceManning:
:+1:  very nice

mmeisinger:
Looks like a very nice expansion of the system.
I cannot not see any issues reading the code. You used the patterns of preload well. I'm sorry if it was a pain to get there. For a dev new on the project, this is a feat that not many accomplished. This is probably one of the most sophisticated pieces of the system. Excellent work!

ion/services/dm/test/test_site_data_products.py
```diff
@@ -0,0 +1,334 @@
...
+from ion.services.dm.test.test_dm_end_2_end import DatasetMonitor
+from ion.services.dm.utility.granule import RecordDictionaryTool
+from nose.plugins.attrib import attr
+
+import numpy as np
+
+import calendar
+from datetime import datetime
+
+from interface.objects import *
```
lukecampbell: importing \* could cause some issues in nosetests.  I don't remember the context but we had issues on buildbot because of import *. It may no longer be an issue though.

ion/services/sa/observatory/observatory_management_service.py
```diff
@@ -628,13 +639,14 @@ def unassign_site_from_deployment(self, site_id='', deployment_id=''):
...
 
-
     def activate_deployment(self, deployment_id='', activate_subscriptions=False):
         """
         Make the devices on this deployment the primary devices for the sites
         """
         #Verify that the deployment exists
         deployment_obj = self.RR2.read(deployment_id)
+        if log.isEnabledFor(logging.DEBUG):
+            log.debug("Activating deployment '%s' (%s)", deployment_obj.name, deployment_id)
```
lukecampbell: I think that it's safe to use log.debug here without checking if it's enabled. 

We do checks for enabled in tight loops to safeguard against being IO-bound in a tight loop. logging, even if it's not enabled may cause gevent to switch contexts and it definitely has overhead.

ion/processes/bootstrap/ion_loader.py
```diff
@@ -3469,6 +3483,44 @@ def update_data_product(dp_id, newrow, const_id1, const_id2):
...
+        # setup SideDataProduct rows based on their stream_configurations column
+        newrows = []
+        for obj in self.resource_objs.values():
+            if obj.type_ in [RT.PlatformSite, RT.InstrumentSite]:
+                if hasattr(obj, "stream_configurations") and obj.stream_configurations:
+                    for index, scfg in enumerate(obj.stream_configurations):
+                        sdp_id = obj._id + "_SDP" + str(index)
+                        newrow = {}
+                        newrow[COL_ID] = sdp_id
+                        newrow['dp/name'] = "SiteDataProduct %s on site '%s' with stream '%s'" % (sdp_id, obj._id, scfg.stream_name)
```
lukecampbell, line 3528: This is fine for now, but we'll need to have a better naming convention in the future. I think that if the UI and ERDDAP all show
- SiteDataProduct X on site Y

I would expect users to complain.

ion/processes/bootstrap/ion_loader.py
```diff
@@ -3469,6 +3483,44 @@ def update_data_product(dp_id, newrow, const_id1, const_id2):
...
+                        newrow['dp/description'] = "SiteDataProduct %s on site '%s' with stream '%s'" % (sdp_id, obj._id, scfg.stream_name)
+                        newrow['dp/ooi_product_name'] = ""
+                        newrow['dp/processing_level_code'] = "N/A"
+                        newrow['dp/quality_control_level'] = "N/A"
+                        newrow['dp/category'] = DataProductTypeEnum._str_map[DataProductTypeEnum.SITE]
```
lukecampbell, line 3533: I didn't even know enums had the _str_map.... damn, that could have saved me so much trouble.

ion/processes/bootstrap/ion_loader.py
```diff
@@ -118,7 +118,7 @@
 TESTED_DOC =  "https://docs.google.com/spreadsheet/pub?key=0AgjFgozf2vG6dGZ6TXdQZ2VTT0phdXMyU0JydmE2cHc&output=xls"
 
 ### while working on changes to the google doc, use this to run test_loader.py against the master spreadsheet
-#TESTED_DOC=MASTER_DOC
+TESTED_DOC=MASTER_DOC
```
lukecampbell: I don't think we can merge safely with this line uncommented.

lukecampbell:
Looks really good, some minor tweaks and I should be able to make more progress.

PR #2062, opened 2014-05-02
Skips for PYCC mode

lukecampbell:


lukecampbell:
@jamie-cyber1 this should fix some DM test failures.

PR #2063, opened 2014-05-05
Fixes test_sparse_values on pycc

lukecampbell:
- The test publishes to a dataset 4 times but the monitor only listened
  to three of them before making assertions causing it to sporadically
  fail. The solution, here, was to add another wait at the third publish.


lukecampbell:
@jamie-cyber1 This one has been failing for far too long, sorry about that.

jamie-cyber1:
:+1:

PR #2066, opened 2014-05-06
CIDEVSTC-84	 Implement notification delivery by rule

MauriceManning:
Implement batch processing for UNS to meet R3 scope. Aligns with updated notifications.yml.
- supports multiple delivery configurations per NR
- adds NR name to event summary in email
- supports notification types: simple, platform, site and facility using outil
- batch processing does not support digest DeliveryMode 

this PR will NOT pass buildbot until the new notifications.yml is merged:
https://github.com/ooici/ion-definitions/pull/564

@mmeisinger please review
@brianmckenna please review
@seman  please review

### DIFF
--- a/ion/services/dm/presentation/user_notification_service.py
+++ b/ion/services/dm/presentation/user_notification_service.py
@@ -369,7 +369,8 @@ class UserNotificationService(BaseUserNotificationService):
 
         @retval notifications list of NotificationRequest objects
         """
-        notifications = []
+        #todo remove and use self.user_id_to_nr_map
+        self.notifications = []
         user_notif_req_objs, _ = self.clients.resource_registry.find_objects(
             subject=user_info_id, predicate=PRED.hasNotification, object_type=RT.NotificationRequest, id_only=False)
 
@@ -378,9 +379,9 @@ class UserNotificationService(BaseUserNotificationService):
         for notif in user_notif_req_objs:
             # do not include notifications that have expired
             if notif.temporal_bounds.end_datetime == '':
-                    notifications.append(notif)
+                    self.notifications.append(notif)
 
-        return notifications
+        return self.notifications
 
 
     def create_worker(self, number_of_workers=1):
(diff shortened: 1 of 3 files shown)


ion/services/dm/presentation/user_notification_service.py
```diff
@@ -426,68 +441,300 @@ def process_batch(self, start_time = '', end_time = ''):
...
+        # retrieve all users in the system
+        users, _ = self.clients.resource_registry.find_resources(restype=RT.UserInfo)
 
-            notifications = self.get_user_notifications(user_info_id=user_id)
-            notifications_disabled = value['notifications_disabled']
-            notifications_daily_digest = value['notifications_daily_digest']
+        for user in users:
+
+            #retrieve all the active notifications assoc to this user
+            notifications = self._get_user_notifications(user_info_id=user._id)
```
mmeisinger: This is not effective database-wise and I suggest to change the loop.
Currently it makes a read for every user in the system, no matter if there are subscriptions or not.
It may be better to initially read all the NotificationRequest resources and all associations at once and then bucket in memory.

ion/services/dm/presentation/user_notification_service.py
```diff
@@ -426,68 +441,300 @@ def process_batch(self, start_time = '', end_time = ''):
...
+
+        # the most common subscription is an origin and an event type so create a map that has that pair as a tuple
+        if origin_id and origin_id != '*':
+
+            # get the union of origins and event types to build the query
+            if not origin_id in self.origins :
+                self.origins.append(origin_id)
+
+            if notification.event_type and notification.event_type != '*':
+                tuple = (origin_id, notification.event_type)
```
mmeisinger: tuple is a reserved word in Python for the type and should not be used as a variable

mmeisinger:
Nice implementation. Good forward progress here!
A couple smaller issues above

MauriceManning:
https://github.com/ooici/coi-services/pull/2070

PR #2071, opened 2014-05-07
Deprecates old QC tests (new ones will replace)

lukecampbell:


lukecampbell:
@jamie-cyber1 ready to merge

jamie-cyber1:
Why not just remove it?

lukecampbell:
Because, once M088 is ready the test module here will need to be refactored. But in the transition period it's deprecated.

PR #2073, opened 2014-05-07
M101 Mission Executive

Bobfrat:
@carueda, @edwardhunter same as before but rebased onto master

-remove print statements
-add mission thread abort functionality


carueda:
Good to be merged, thanks!

PR #2093, opened 2014-05-13
OOIION-1866 - Updates metadata when parameters are added to a data product

lukecampbell:


lukecampbell:
@jamie-cyber1 Take a look please.  I added a new dependency beautifulsoup4, the EOI team needs it as well, and it's probably the best XML parsing module I've laid eyes on since JQuery.

lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/251

jamie-cyber1:
Looks good!  You might want to think about speeding up the reply from the lone register dataset process (if you cannot create multiple replicas) so it will not become a bottleneck.  One such async example would be process dispatcher's schedule_process call.

PR #2100, opened 2014-05-14
OOIION-1856: Fixed bug in policy function for activate/deactivate deployment

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #2111, opened 2014-05-15
Bump iondef to include production preload yaml for R3

jamie-cyber1:


jamie-cyber1:
@mmeisinger please merge.

PR #2115, opened 2014-05-16
OOIION-1855: Set site/deployment/model attributes in extended res

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge (after ion-defs)

mmeisinger:
@jamie-cyber1 ion-def pointer updated, ready to go after test

PR #2130, opened 2014-05-22
CIDEVSTC-99 - M112 - support search radius in meters

brianmckenna:
Adds tests for radius specified in meters and cleaned up leftover print statements

@mmeisinger please review for merge, only tests and pyon bump


mmeisinger:
:+1:

PR #2133, opened 2014-05-28
CISWMI-222: Fix issue with the generation of DP for cabled inst

mmeisinger:


mmeisinger:
@jamie-cyber1 please merge

PR #2135, opened 2014-05-28
TOOLS: OOI preload enhancements - names, urls, clones

mmeisinger:
- Fix name of instruments to include makemodel and SAF class-series
- Add DPS URLs to DataProducts
- Ability to clone nodes and instruments in OOI preload for bench sites and additional gliders


mmeisinger:
@jamie-cyber1 please merge

PR #2141, opened 2014-05-30
M101 Add events for starting and completing missionSequences

Bobfrat:
-Also add an example shallow profiler mission for RSN
@carueda, @edwardhunter please review


carueda:
Looks good. But later on we can on decide if/how to more properly differentiate these new events from regular "thread started/ended" events.

PR #2138, opened 2014-05-29
M101 - add description to mission executive events

Bobfrat:
Also clean up the code a bit

@carueda @edwardhunter please review

PR #2146, opened 2014-06-02
M101: exclusive access for mission execution

carueda:
As suggested, the exclusive access request is based on the ORG.acquire_resource call (and not via the negotiate call). However, there's some ad hoc mechanism to get a negotiation_id that is required by the acquire_resource call. More details in the code comments included in this PR.

@mmeisinger / @edwardhunter please review.

(this PR replaces #2138, which was taking me much time trying to update)


carueda:
In summary:
- in agent_configuration_builder:
  - the Org associated with the agent instance is retrieved and its ._id value is captured as entry 'provider_id' in the agent config
  - a set_actor_id method was added; corresponding value is captured as entry 'actor_id' in the agent config
- in IMS, the actor_id is retrieved from the context ('ion-actor-id') and set it into the configuration builder
- in base_test_platform_agent_with_rsn, the value returned by get_system_actor_header is used as headers argument for IMS.start_platform_agent_instance call
- in platform_agent/mission_manager, from passed CFG 'provider_id' and 'actor_id' are used for 'provider' and 'consumer' attributes of AcquireResourceExclusiveProposal object for the ORG.acquire_resource call.

Some of the pending aspects:
- a negotiation_id argument is required by ORG.acquire_resource; this id is built with an RR call
- ORG.acquire_resource also requires a headers argument, which is now constructed based on get_system_actor_header; but what's the correct way to obtain the actor header?
- expiration for the AcquireResourceExclusiveProposal: now with an ad hoc value; Could it be without expiration given that the exclusive access will be removed explicitly upon termination (normal or abnormal) of the the mission?

ion/agents/platform/mission_manager.py
```diff
@@ -51,21 +80,30 @@ def run_mission(self, mission_id, mission_yml):
...
             log.exception('[mm] run_mission mission_id=%r', mission_id)
         finally:
             del self._running_missions[mission_id]
+
+            # remove exclusive access:
+            mission_entries = mission_loader.mission_entries
+            for mission_entry in mission_entries:
+                instrument_ids = mission_entry.get('instrument_id', [])
+                for instrument_id in instrument_ids:
+                    self._remove_exclusive_access(instrument_id, mission_id)
```
mmeisinger, line 105: I wonder if anything could fail here and you would be better off wrapping each such call in the for loop in an exception handler. Otherwise you may fail halfway in between. Of course you still need some form of compensation logic, but at least you try to execute on every instrument

ion/agents/platform/mission_manager.py
```diff
@@ -136,9 +179,180 @@ def _create_mission_scheduler(self, mission_id, mission_yml):
...
+        # we are initially opting for only "phase 2" -- just acquire_resource:
+        commitment_id = self.ORG.acquire_resource(arxp, headers=self._actor_header)
+        log.debug('[xa] AcquireResourceExclusiveProposal: '
+                  'resource_id=%s -> commitment_id=%s', resource_id, commitment_id)
+        return commitment_id
+
+        # #####################################################################
+        # # with "negotiation" it seems it would involve something like the
+        # # following (see coi/../test_gobernance.py):
+        # # 1- negotiate a base acquire resource proposal
```
mmeisinger, line 297: just a comment: It can be expected that a user starting a missing is either a high level manager not needing acquire resource permissions, or is a facility operator that already has acquired all instrument resources. I would not expect this call to do the acquire resource negotiations here. If they don't exist, the mission cannot be started and the user needs to be notified to acquire the missing resources first.
Also note that acquire resource is a human in the loop process and probably takes hours to days depending on how quickly a manager approves.

mmeisinger:
Very nicely done! :+1: 
I like the extension of the agent config builder and the code to request/remove exclusive access in the agent.

Note one weakness: The actor starting the agent may not be the actor starting the mission. We may start agents using system level scripts with no actor. Once started, any actor may actually start a mission.
You basically inherit the actor from the start agent call. It would be best to get the current actor out of the RPC call to the agent, IMO. This can be discussed.

ion/agents/platform/mission_manager.py
carueda, line 105: Agree. and note that I'm relying on the fact that _remove_exclusive_access catches any exception (currently just to log it out) so it doesn't prevent this loop to fully complete.

carueda:
Thanks Michael.
How "to get the current actor out of the RPC call to the agent"? Perhaps it's a straightforward call that I'm missing to see and that I could add for this PR..

mmeisinger:
The code to get the actor header is within an RPC:

# typically

process = self
ctx = process.get_context()
actor_id = ctx.get('ion-actor-id', None) if ctx else None

It's only set in the greenlet that executes the RPC call before any response has been sent. I would assume this is true for the agent as well.

ion/agents/platform/mission_manager.py
```diff
@@ -35,6 +40,30 @@ def __init__(self, pa):
         # mission_id -> MissionScheduler mapping:
         self._running_missions = {}
 
+        self._provider_id = self._agent._provider_id
+        self._actor_id = self._agent._actor_id
+        log.debug('[xa] provider_id=%r  actor_id=%r', self._provider_id, self._actor_id)
+
+        # ctx = self._agent.get_context()
+        # self._actor_id = ctx.get('ion-actor-id', None) if ctx else None
+        # log.debug('[xa] actor_id=%r', self._actor_id)
```
carueda, line 49: Re your previous comment on the main thread, I think that's what I tried here (now commented out), but always getting an empty string so maybe I got confused and started looking into retrieving the actor_if by other means.  So, would this commented out code the actual mechanism to retrieve the actor_id? IF so, what would be the way to do the corresponding preparation in the integration test set-up to exercise this?

mmeisinger, line 49: Yes this seems correct. self._agent needs to be the process of the agent. Of course the test needs to make the agent call with the ion-actor-id header set.

This code must be called within the same greenlet that executes the RPC. Otherwise the call may not error out but may not find the message headers.

ion/agents/platform/test/base_test_platform_agent_with_rsn.py
```diff
@@ -1392,7 +1395,8 @@ def consume_event(evt, *args, **kwargs):
         # now start the platform:
         agent_instance_id = p_obj.platform_agent_instance_id
         log.debug("about to call start_platform_agent_instance with id=%s", agent_instance_id)
-        p_obj.pid = self.IMS.start_platform_agent_instance(platform_agent_instance_id=agent_instance_id)
+        p_obj.pid = self.IMS.start_platform_agent_instance(platform_agent_instance_id=agent_instance_id,
+                                                           headers=self._actor_header)
```
carueda, line 1399: Which seems to be what we have here, right? If so, perhaps I tried various combinations but not precisely one with this call and the commented out on the other file ... which I just tried it but got None from self._agent.get_context().

When you say "self._agent needs to be the process of the agent," is that a different thing and not the agent itself?

mmeisinger:
:+1: ready to go

PR #2152, opened 2014-06-03
M101 - Fix event driven mission logic

Bobfrat:
@carueda, please review
@edwardhunter please merge

### DIFF
--- a/ion/agents/mission_executive.py
+++ b/ion/agents/mission_executive.py
@@ -17,9 +17,9 @@ from pyon.agent.agent import ResourceAgentEvent
 from pyon.agent.common import BaseEnum
 from pyon.event.event import EventSubscriber
 from pyon.public import log
+from pyon.util.breakpoint import breakpoint
 from pyon.util.config import Config
 
-from ion.agents.platform.platform_driver_event import ExternalEventDriverEvent
 from ion.agents.platform.rsn.rsn_platform_driver import RSNPlatformDriverEvent
 from ion.core.includes.mi import DriverEvent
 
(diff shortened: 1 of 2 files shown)


ion/agents/mission_executive.py
```diff
@@ -918,25 +918,28 @@ def _run_event_driven_mission(self, mission):
...
-            origin = self.resource_id
+            origin = self.platform_agent._platform_id
         else:
-            self._publish_mission_thread_failed_event(current_mission_thread, 'Parent ID unavailable')
-            raise Exception('Parent ID unavailable')
+            self._publish_mission_thread_failed_event(current_mission_thread, 'Parent ID unrecognized - {0}'.format(parent_id))
+            raise Exception('Parent ID unrecognized - {0}'.format(parent_id))
 
         # Check that the event id is legitimate
         if event_id in DriverEvent.__dict__.keys():
```
carueda, line 909: This is in the body of _run_event_driven_mission, so I understand event_id is the event ID indicated in the mission plan, meaning that when that asynchronous event is received, the execution of the corresponding sequences is triggered.

Now, DriverEvent is an enum with "events for driver state machines", so they are not events in the sense of those objects that are published somewhere and asynchronously received by subscribers. (Note that the same "event" term is used in the FSM context, but there they would probably be better called "commands" or "inputs" instead.)

ion/agents/mission_executive.py
```diff
@@ -918,25 +918,28 @@ def _run_event_driven_mission(self, mission):
...
             event_type = 'ResourceAgentCommandEvent'
             event_id = getattr(DriverEvent, event_id)
+        elif event_id in RSNPlatformDriverEvent.__dict__.keys():
```
carueda, line 912: Ditto

ion/agents/mission_executive.py
```diff
@@ -918,25 +918,28 @@ def _run_event_driven_mission(self, mission):
...
-        elif event_id in RSNPlatformDriverEvent.__dict__.keys():
-            event_type = 'OMSDeviceStatusEvent'
-            event_id = getattr(ExternalEventDriverEvent, event_id)
         else:
-            self._publish_mission_thread_failed_event(current_mission_thread, 'Event ID unavailable')
-            raise Exception('Event ID unavailable')
+            # EXTERNAL event - check OMSDeviceStatusEvent
+            event_type = 'OMSDeviceStatusEvent'
+            #self._publish_mission_thread_failed_event(current_mission_thread, 'Event ID unrecognized - {0}'.format(event_id))
+            #raise Exception('Event ID unrecognized - %s', event_id)
```
carueda, line 920: Perhaps there's not even need for the "if event_id in MissionEvents.**dict**.keys()" case: rather, there would only be the else here for the "external event" case.  This would assume that any explicitly unrecognized event_id from the mission plan would be assumed to be an external event (to be received with an OMSDeviceStatusEvent).

Bobfrat, line 920: New commit omits the entire MissionEvents class from mission_executive.py. It was not needed there as it was only used to test a simulated profiler.

carueda:
Thanks Bob.
@edwardhunter Assuming tests are running as usual, please merge.

carueda:
Bob: please add the missing import.

Import module 'ion.agents.platform.test.test_mission_executive' failed: name 'BaseEnum' is not defined

ion/agents/platform/test/test_mission_executive.py
```diff
@@ -68,6 +68,21 @@ def get_agent_state(self):
         return "FAKE"
 
 
+class MissionEvents(BaseEnum):
```
carueda, line 72: so, can you either also omit it here or import the missing BaseEnum?

Bobfrat:
import issue fixed

carueda:
Thanks Bob.
@edwardhunter Assuming tests are running as usual, please merge.

PR #2153, opened 2014-06-03
minor refactor in platform driver's get_resource_capabilities

carueda:
with new option to retrieve actual capability enum class attributes (instead of the corresp values, which continues to be the default behaviour). This will help make the mission executive more general.

@edwardhunter please merge

@bobfrat :  You can then call:

``` python
    res_cmds, res_params = self.platform_agent._plat_driver.get_resource_capabilities(cmd_attrs=True)
```

and `res_cmds` will be the actual class attributes, so this will help with your scheme of things in the mission executive.

### DIFF
--- a/ion/agents/platform/platform_driver.py
+++ b/ion/agents/platform/platform_driver.py
@@ -16,7 +16,6 @@ import logging
 
 from copy import deepcopy
 
-from ion.agents.platform.platform_driver_event import DriverEvent
 from ion.agents.platform.platform_driver_event import StateChangeDriverEvent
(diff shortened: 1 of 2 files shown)


carueda:
Do not merge yet!  THis still needs to be completed.

carueda:
Done. What was missing was an operation to get the actual platform event class.

So, with a retrieved res_cmds above, say cmd = 'TURN_ON_PORT', you can then look up the corresponding value in the platform driver:

``` python
driver_event_class = self.platform_agent._plat_driver.get_platform_driver_event_class()
command = getattr(driver_event_class, cmd)
```

PR #2163, opened 2014-06-06
M105 Notification Personalization and Major Code Cleanup

lukecampbell:
As you can see, I removed a lot of unused code.
- Workflow Management Service
- Preservation Management Service
- Unused Transforms
- Tests

All I did was look for modules that weren't referenced, these shouldn't impact anything.


lukecampbell:
@jamie-cyber1 @MauriceManning @mmeisinger We'll wait for coi_pull but should be good to go. Please look at the file listing to make sure I'm not removing anything that you might want to use in the near future for R3.

mmeisinger:
I looked through all files and there is nothing that I would like to keep. I am glad the old stuff will be gone. Makes refactorings and searches in the code base so much more difficult.
@MauriceManning are all the event transforms OBE and subsumed in the new processing framework, or just not needed anymore?

lukecampbell:
I wondered about that very thing but the event transform stuff is in a different class and tuned for R3s data processing work from M087.  These weren't referenced at all based on what grep said.

PR #2169, opened 2014-06-09
M101: Add assertions to mission_manager tests

Bobfrat:
@carueda, @edwardhunter,  please review


carueda:
Looks good.
@edwardhunter please merge.

Quick general comment that _start_event_subscriber2 is being called with count=1, so, in a strict manner, comparing the number of received events with 6 is not robust enough because only the first event makes the wait to complete. What happens now is that we have some interleaving logic that actually waits until the mission execution completes. So, by the time the comparison to 6 is done, those events would have been received as well.  So, not critical, but something to keep in mind.

PR #2174, opened 2014-06-10
Removes passing coverage objects to the EOI handler

lukecampbell:


lukecampbell:
@birdage  please review

birdage:
@lukecampbell need fill value, but other than that :+1:

PR #2176, opened 2014-06-11
M101: initial version of instrument stable ids based on PRE:* entries in alt_ids

carueda:
@edwardhunter / @bobfrat please review 

worth noting:
- For integration test purposes, according to our keys for the SBE simulators,  I'm using alt_ids of the form "PRE:SBE_", that is, using the prefix "PRE:" (as suggested to appear in real alt_ids).  The suffix can be anything but I'm using the same SBE37_ just to easily correlate visually in the logs.
- All YAML files that we are testing at the moment haven been updated. Note that per YAML syntax rules, it's  necessary to enclose those IDs (because of the colon) in quotes in some places.
- Those YAML files are now loaded as given (no more replacement to resource ids)
- Focus now is on the integration tests -- nothing yet done on ion_loader or similar
- all tests completing OK

Please review.

### DIFF
--- a/ion/agents/mission_executive.py
+++ b/ion/agents/mission_executive.py
@@ -620,9 +620,10 @@ class MissionScheduler(object):
         log.debug('[mm] schedule: waiting for mission to complete')
         gevent.joinall(self.threads)
 
-    def _send_command(self, agent_client, cmd):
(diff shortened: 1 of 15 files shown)


Bobfrat:
Looks good assuming the commands in the YAML work with the colon in the instrument id.

PR #2182, opened 2014-06-13
M057: update Dataset Management Service and Generation of URL from external data products 

birdage:
a number of updates to dataset management service
- add private data verification
- removes duplicate code from update dataset

updates DPMS url generation for EOI data products (verified against the flag)

fixes a few bugs in table loader


birdage:
@jamie-cyber1 @mmeisinger Please review

lukecampbell:
:+1:

PR #2184, opened 2014-06-13
More cleanup and speed up some tests

lukecampbell:


lukecampbell:
@MauriceManning @mmeisinger Can you please review and make sure I'm not removing any modules you want to keep?

mmeisinger:
:+1: nothing i would like to keep

PR #2192, opened 2014-06-17
OOIION-1920: Enhanced support for UI search

mmeisinger:


mmeisinger:
merge ion-definitions first

PR #2193, opened 2014-06-18
Removes assert from test

lukecampbell:
- We no longer raise, we just ignore it


lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/260

PR #2194, opened 2014-06-18
R3 Coverage Model Integration

lukecampbell:
- Initial Integration for support on M193

Outstanding:
- Further Array-type support (it works completely in all aspects except we want to add the ability to learn an inner-dimension size at the first ingestion so that we don't have to update 360+ array type parameter definitions and research the inner array size for each one)
- Overwrite values (M088)
- External Function types (M086)


lukecampbell:
@MauriceManning @mmeisinger @birdage @caseybryant
Ready for review and discussion.

lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/261

lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/262

PR #2195, opened 2014-06-18
Platform tests sbe37sim

carueda:
@jamie-cyber1 / @wbollenbacher / @edwardhunter please review/merge.

In short, only the simulator on port 4001 (as indicated by CFG.device.sbe37.port) is now used.

Note: the DATA_PORT and CMD_PORT parameters for the various instruments (eg, 5001/6001, 5002/6002, etc) continue to be defined in the base test class as only a single pair of those for all instruments launched concurrently would conflict.  But the core request here to only use the simulator on 4001 is now in place. (And very nice that that single simulator can be instantiated multiple times.)

BTW, perhaps it's a good time to think again about capturing all of these parameters in pyon.yml as proposed long time ago.


carueda:
@mmeisinger perhaps you can go ahead and review/merge?

mmeisinger:
:+1:

PR #2198, opened 2014-06-23
Resumes array type testing

lukecampbell:
- Bumps the cov-model to include latest array type support
- Un-skips the tests and resumes testing.


lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/264

PR #2212, opened 2014-07-02
Fixes test_serialize_compatibility

lukecampbell:
- Increases timeout in test_serialize_compatability


lukecampbell:
http://buildbot.oceanobservatories.org:8010/builders/coi_any/builds/277

PR #2230, opened 2014-07-18
OOIION-1915 Mission Files Time Format String

Bobfrat:
-Change startTime to ISO-8601
-Keeps current timezone handling though


Bobfrat:
@edwardhunter please review
@carueda FYI...

